Issue #25, opened 2025-08-19
ghc-debugger shouldn't consult "ghc" on PATH


mpickering:
Currently `ghc-debugger` attempts to determine the version that will be used by looking at the "ghc" on PATH.

This is incorrect because the version selected may be different from the version on PATH (ie if you use `with-compiler` in a cabal project file)

The correct API to use is [getRuntimeGhcVersion](https://hackage-content.haskell.org/package/hie-bios-0.17.0/docs/HIE-Bios-Environment.html#v:getRuntimeGhcVersion) I believe.

PR #26, opened 2025-08-19
extension: Look for -<version> suffixed executable

alt-romes:
Matching the install instructions (on website) which suggest setting program-suffix:

```
cabal install ghc-debugger:ghc-debug-adapter --program-suffix="-$(ghc --numeric-version)"
```

### DIFF
--- a/vscode-extension/src/extension.ts
+++ b/vscode-extension/src/extension.ts
@@ -33,6 +33,33 @@ function getFreePort(): Promise<number> {
 	});
 }
 
+function spawnDebugger(logger: vscode.OutputChannel, port: number) {
+    let adapterCmd = 'ghc-debug-adapter';
+
+    try {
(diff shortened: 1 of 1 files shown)


alt-romes:
I suspect this may be at odds with #25 ...

I thought it would be good to have the versions of the debugger be installed suffixed by the ghc version they were compiled with. We'd look up the debugger executable by the version of the GHC in PATH. This would help avoid using the wrong version of the debugger.

But if the debugger should not look at the GHC in PATH, it's harder to do this kind of validation.

alt-romes:
Perhaps it ought to be a property of the `launch.json` file which GHC version to use?
In that case, every launch would use the predefined GHC version to suffix the debugger executable to spawn. By default the variable could be configured to something like $GHC_IN_PATH.

@mpickering thoughts?

PR #59, opened 2025-09-13
Don't display functions as unevaluated thunks

### LINKED ISSUE #48: Variables: Don't display functions as unevaluated thunks (_)
In the variable pane:

Image

The local binding  `versionOr` is a function. However, it is displayed as an unevaluated thunk `"_"` in the pane.

It would be good to do something else in the case of functions and friends (such as DataCons), for instance, to display its type inline.

amaebel:
This pr adds the type of functions to the values shown in the variables view to provide useful information instead of always displaying functions as unevaluated thunks.

closes #48


alt-romes:
Looks great!

PR #62, opened 2025-09-14
fix: changed  default config settings for missing launch.json case.

LordN909:
Changes the default configuration to start the debugger if launch.json does not exist.


alt-romes:
Fixes #63

Issue #47, opened 2025-09-03
hdb: Improve printing for some datatypes


alt-romes:
Some datatypes should be pretty printed by default rather than using the internal memory representation.

Some which come to mind:

- [ ] ByteString
- [ ] Text
- [x] String
- [ ] IntMap
- [ ] Maps
- [ ] other containers...

meipp:
Currently, I check for instances of the class by unqualified names like this:

```hs
isDebugInstance :: Term -> Type -> Debugger Bool
isDebugInstance term t = do
  hsc_env <- getSession

  instances <- getInstancesForType t
  let classNames = map (occNameString . occName . tyConName . classTyCon . is_cls) instances

  case filter ((== "Debug") . occNameString . occName . tyConName . classTyCon . is_cls) instances of
    (c:_) -> do
      let methods = (classOpItems . is_cls) c
      traceM ("Found Debug instance with methods: " ++ (show . map (occNameString . occName . fst)) methods ++ "")
      case filter ((== "debugDisplayTree") . occNameString . occName . fst) methods of
        (m:_) -> do

          let dfun = is_dfun c
          traceM $ "Dictionary function: " ++ showSDocUnsafe (ppr dfun) ++ " :: " ++ showSDocUnsafe (ppr (varType dfun))

          let method_id = fst m :: Id
          traceM $ "debugDisplayTree method: " ++ showSDocUnsafe (ppr method_id) ++ " :: "  ++ showSDocUnsafe (ppr (varType method_id))

          (method_hv, _, _) <- liftIO $ Loader.loadName (hscInterp hsc_env) hsc_env (GHC.Types.Var.varName method_id)
          (dfun_hv, _, _) <- liftIO $ Loader.loadName (hscInterp hsc_env) hsc_env (GHC.Types.Var.varName dfun)
          return True
        [] -> return False
      return False
    _  -> return False
```

meipp:
Not sure how to work with `method_hv` and `dfun_hv` from here on.

```hs
ev <- liftIO $ evalStmt (hscInterp hsc_env) (initEvalOpts (hsc_dflags hsc_env) EvalStepNone) (EvalApp (EvalThis dfun_hv) (EvalThis method_hv))
```
and calling it with `method_hv` and `dfun_hv` flipped
```hs
[...]

PR #75, opened 2025-09-19
Incomplete state of Debug instance visualization

meipp:
Addresses #47


alt-romes:
I've rebased this on master. Looking into the segfault now.

alt-romes:
@meipp what is your status regarding your work? If you'd like, I can pick this up and finish it.

The main difficulty is designing the type class, but I've got a few ideas to try. The challenging part of calling the instance dynamically you've already figured out, so I can just re-use that for a custom class.

This is now more of a priority for the project I believe and I'd be willing/looking forward to work on this now

PR #125, opened 2025-11-17
Add a cache for `DebugViewInstances`

### LINKED ISSUE #122: Cache DebugViewInstances per type
Follow up to #120 : Create a TypeMap cache which enables us to re-use the found instances instead of recompiling and finding them again.

alt-romes:
It allows us to avoid a lot of repetitive work compiling a method and loading and running it.

I confirmed locally that these instances were now only being searched for and compiled once.

Fixes #122


alt-romes:
It's harder to test in CI since the log for compiling an instance is only produced in DEBUG verbosity.

PR #148, opened 2025-12-03
Windows fixes and CI

### LINKED ISSUE #146: unix dependency prevents use on Windows
We depend on `unix` currently for `System.Posix.Signals`. A dependency on `unix` will prevent haskell-debugger from being used on Windows, where `unix` is not available.

Don't depend unconditionally on unix!!

### LINKED ISSUE #147: CI: Build and test haskell-debugger on Windows
https://discourse.haskell.org/t/installation-of-newer-versions-of-stack-with-server-on-visual-studio-code/13322/10 uncovered a build failure on windows caused by #146.

We should make sure we at least try to build `haskell-debugger` on Windows in CI, to make sure this doesn't regress. We should run the tests too.

alt-romes:
Fixes #146 
Fixes #147

### DIFF
--- a/.github/workflows/debugger.yaml
+++ b/.github/workflows/debugger.yaml
@@ -25,7 +25,7 @@ name: Debugger CI
 jobs:
   build-vscode-extension:
     name: Build VSCode Extension
-    runs-on: ubuntu-latest # or macOS-latest, or windows-latest
+    runs-on: ubuntu-latest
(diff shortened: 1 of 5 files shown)


alt-romes:
and the DAP proxy test does file handle things which don't seem to work on windows. Let's disable that test on Windows for now.

PR #165, opened 2026-01-14
Improvements to exceptions

mpickering:
Note, requires https://github.com/haskell-debugger/dap/pull/26

### DIFF
--- a/haskell-debugger.cabal
+++ b/haskell-debugger.cabal
@@ -61,6 +61,7 @@ library
                       GHC.Debugger.Logger,
                       GHC.Debugger.Run,
                       GHC.Debugger.Stopped,
+                      GHC.Debugger.Stopped.Exception,
                       GHC.Debugger.Stopped.Variables,
 
                       GHC.Debugger.Runtime,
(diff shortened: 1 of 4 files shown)


alt-romes:
Brilliant stuff. The MR looks great. Just a matter of fixing the CI

mpickering:
@alt-romes This is ready for review now.

alt-romes (changes requested):
This looks great.

I commented on some individual commits but I see that in the end some things ended up different, so you can ignore my previous comments in favor of the following two requested changes:

- Move the whole new section `"Exception context helpers"` to a new module `GHC.Debugger.Stopped.Exception`
- Use MultilineStrings for the long raw expressions (`exceptionInfoExpr` and `exceptionLocationExpr`)

Great work!

mpickering:
I moved the helpers to a new module in a new commit.

mpickering:
For the record, Rodrigo says that the reason there are 5 exceptions thrown in the integration test is that..

1. First call to `error`
2. Second call to `error`
3. Rethrow from using `catch`
4. Rethrow from using `bracket` in `withProgName` called in `evalWrapper`
5. Rethrow from using `bracket` in `withArgs` called in `evalWrapper`

PR #195, opened 2026-02-26
Display all threads by default

alt-romes:
- We display all the threads in the debuggee process, except for ...
- When using the internal interpreter, the debuggee process and debugger process are the same. Therefore, we'll also see the debugger threads in that case. That's fine, it's observing the actual threads of the process being debugged.


alt-romes:
closed in favor of #371

PR #197, opened 2026-02-28
Remove unnecessary periods from cabal project description

cheater:
The periods show up as double periods on Hackage:

image


alt-romes:
Thanks!

PR #203, opened 2026-03-04
Add cabal to build-tool-depends

alexfmpe:
Needed due to https://github.com/well-typed/haskell-debugger/blob/d3da6051d1f9ab08fdf4907c28f46af5e4236270/test/golden/T79/T79.hdb-test#L1

as it causes 
```
T79:                          T79.hdb-test: line 1: cabal: command not found
```
when building in contexts where `cabal-install` can not be assumed to be in PATH.


alt-romes:
Nicely spotted! Thanks

PR #223, opened 2026-03-17
[Fixes #113] LogMessage support

### LINKED ISSUE #113: Support Log Points (supportsLogPoints capability)
Akin to #111, DAP supports setting log points: when a log point breakpoint is hit we simply evaluate and log the given expression and do not stop the execution.

Saizan:
Got it to work, including the interpolation.

TODO:
- [x] CLI support
- [x] tests

I also left a couple of FIXME for places where I kept the old code behaviour but I wasn't sure if that was the right thing.


alt-romes (commented):
Thanks! The design is good, just a few things to tweak, the testsuite to fix, and a new test for log points is needed.

You can ignore the CLI mode for now.

alt-romes (approved):
Looks perfect! Did you enable `supportsLogPoints` in the initialization message?

alt-romes (approved):
Excellent work! Let's land it

PR #235, opened 2026-03-25
Fix #234 T130b

### LINKED ISSUE #234: Custom debug view instances do not work on newtypes
A `DebugView` instance like:
```hs
instance View.DebugView UnitId where
  debugValue uid = View.VarValue (pure (unitIdString uid)) False
  debugFields _ = pure (View.VarFields [])
```

Currently doesn't work because `UnitId` is a new type. It's probably easy to construct a standalone example where the debug view instance doesn't show the right result when printing the value because the instance is given for a new type.

It's easy to see this is wrong in the code:

```hs
debugValueTerm :: Term -> Debugger (Maybe VarValueResult)
debugValueTerm term@(Suspension{} ; Term{}) = do
   ...
```

```hs
debugValueTerm :: Term -> Debugger (Maybe VarValueResult)
debugValueTerm term@(Suspension{} ; Term{}) = do
   ...
debugValueTerm term = do
  logSDoc Logger.Warning (text "debugValueTerm: Expecting" <+> ppr term <+> text "to be a Term or Suspension.")
  return Nothing


debugFieldsTerm :: Term -> Debugger (Maybe [(String, Term)])
debugFieldsTerm term@(Suspension{} ; Term{}) = do
  ...
debugFieldsTerm term = do
  logSDoc Logger.Warning (text "debugValueTerm: Expecting" <+> ppr term <+> text "to be a Term or Suspension.")
  return Nothing
```

We must support custom debug views for newtypes too!

alt-romes:
Trying to fix #234, potentially #238 at the same time

### DIFF
--- a/haskell-debugger/GHC/Debugger/Monad.hs
+++ b/haskell-debugger/GHC/Debugger/Monad.hs
@@ -37,6 +37,7 @@ import qualified Data.List as L
 import qualified Data.List.NonEmpty as NonEmpty
 
 import GHC
+import GHC.Data.FastString
 import GHC.Data.StringBuffer
 import GHC.Driver.Config.Diagnostic
 import GHC.Driver.DynFlags as GHC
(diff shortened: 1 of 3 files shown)


alt-romes:
#281 is a better (actually working) fix.
I'm trying the testsuite on 680cb78895184f3348e836388ed6757170515e39 and may open a new MR.

Issue #217, opened 2026-03-12
Some unexpandable variables are still shown as expandable


alt-romes:
Currently, we will display top-level functions in the variables pane such as `f :: Int -> Int` as `f = <fn> :: Int -> Int`.
We mark symbols we know to be functions as "non-expandable" by returning a `variablesReference = 0` for that symbol:

Example:
```
{-# OPAQUE f #-}
f x = x + 1

main = do
  let doit = const (pure () :: IO ())
  putStrLn "helio"
  doit () >> print (f (345**2))
  putStrLn "sdjifskd"
```
We will return the following for the local variables:
```
{
  body = {
    variables = { {
        evaluateName = "_result",
        name = "_result",
        presentationHint = {
          lazy = true
        },
        type = "IO ()",
        value = "_",
        variablesReference = 5
      }, {
        evaluateName = "doit",
        name = "doit",
        presentationHint = {
          lazy = false
        },
        type = "forall {b}. b -> IO ()",
        value = "<fn> :: forall {b}. b -> IO ()",
        variablesReference = 0
      } }
  },
  command = "variables",
  request_seq = 11,
  seq = 12,
  success = true,
  type = "response"
}
```

On the other hand, `_result`, an `IO ()` action, has `variablesReference = 5` which tells the debug adapter client that it can be further expanded. But it cannot! The IO action is also a function under the hood, and it cannot be inspected.

VSCode is somewhat clever here because, while it displays the variable as expandable and lazy, it won't actually display anything further if we click the little eye button, despite the DAP server returning indeed the same variable:

<img width="719" height="228" alt="Image" src="https://github.com/user-attachments/assets/b61a9f76-8349-4a29-acbe-284918a48e6b" />

The `nvim-dap` + `nvim-dap-view` adaptor is not so clever, so it interprets the same value we incorrectly return as the expansion of the original one, despite still returning them all as lazy:

<img width="1041" height="470" alt="Image" src="https://github.com/user-attachments/assets/6b911a19-08a7-4193-9397-f99dd9cd9d9f" />

## Design

I think the design should be more uniform:

- If something is a function directly or under many layers of newtypes, it should not be displayed as a lazy variable nor expandable
- Instead, place the function signature or new type wrapper type as the signature as the value. (rather than `_`)

## Solution

The check for whether something is a function or not is rather too simplistic at the moment. See `termToVarInfo`:

```haskell
[...]

PR #240, opened 2026-03-26
[Fix #217] Normalise type for function check

Saizan:
GHC has a handy `topNormaliseType_maybe` to normalise away newtypes and type family applications (and get at the representation type), which seems to do the job.

TODOs:
- ~~[ ] Find testcase where we need to go through NewtypeWrap to get at the Suspension (or remove code).~~
   - NewtypeWrap only generated on top of Term atm.
- [x] Add a test with equality constraints: `let foo : forall a. a ~ IO () => a = print ()`
- [x] Refactor call to `getInstFamEnvs'` to only do it once at the top?
- [x] Squash eventual wip commits
- [x] Check Term generation for insight in the type variable case
  - The rtti type of a term can be a type variable even if the term is a non-polymorphic value.
- [x] add test that newtypes of datatypes are displayed as expandable.
- [x] Implement plan at https://github.com/well-typed/haskell-debugger/pull/240#discussion_r3001328904

### DIFF
--- a/haskell-debugger/GHC/Debugger/Stopped/Variables.hs
+++ b/haskell-debugger/GHC/Debugger/Stopped/Variables.hs
@@ -5,6 +5,7 @@
 module GHC.Debugger.Stopped.Variables where
 
 import Control.Monad.Reader
+import qualified Colog.Core as Logger
 
 import GHC
 import GHC.Types.FieldLabel
@@ -12,7 +13,10 @@ import GHC.Types.Id.Info
 import GHC.Types.Var
 import GHC.Runtime.Eval
 import GHC.Core.DataCon
+import GHC.Core.FamInstEnv (topNormaliseType_maybe, emptyFamInstEnvs)
 import GHC.Core.TyCo.Rep
+import GHC.Core.Reduction (Reduction(..))
+import GHC.Tc.Instance.Family (tcGetFamInstEnvs, FamInstEnvs)
 import qualified GHC.Runtime.Debugger     as GHCD
 import qualified GHC.Runtime.Heap.Inspect as GHCI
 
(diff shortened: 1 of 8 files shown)


haskell-debugger/GHC/Debugger/Stopped/Variables.hs
```diff
@@ -182,7 +184,44 @@ termToVarInfo key term0 = do
...
+    -- but should not be shown as thunks in the UI.
+    isFunctionType = \case
+      (FunTy _ _ _ _) -> True
+      (ForAllTy _ t)  -> isFunctionType t
+      (topNormaliseType_maybe fam_envs -> Just Reduction{..})
+        -> isFunctionType reductionReducedType
+      _ -> False
+    -- FIXME: couldn't find test case where this is needed.
+    unwrapNewtype (NewtypeWrap {wrapped_term}) = unwrapNewtype wrapped_term
+    unwrapNewtype x = x
```
alt-romes: I think `unwrapNewtype` might be wrong. Especially since we don't find a test, better to leave it out (if it seemed correct then we'd leave it).

But imagine that we're getting the variable information for a new type constructor wrapping a thunk; like:

```hs
newtype X a = X { x_val :: a }

x = <thunk>

y = X x
```

When you inspect `y`, you expect to see

```
> y = X
```

If you expand by clicking on the `>`, you should see
```
> y = X
    x_val = _
```

I think with your current code calling `unwrapNewtype` before checking for suspension/not; we'd rather end up with:
```
> y = _
```

even though `y` is of type newtype wrapping a thunk.

alt-romes (commented):
Excellent work. Just a few comments.

I think we shouldn't do `unwrapNewtype`. See above

PR #253, opened 2026-03-30
test: Add test for DebugView orphan instances

alt-romes:
Seemingly, #237 is not about all orphan instances, but rather some
    slightly more contrived scenario that prevents us from using a
    `DebugView ModuleGraph` defined as an orphan in the debugger when
    debugging the debugger.

    This test checks that orphan instances in a normal situation work
    properly.

Complements the tests added in the fix to #237


alt-romes:
Nevermind @Saizan, the only issue here is that #251 hadn't landed yet. Now that it has, this will work without further work.

Issue #259, opened 2026-04-02
Use `runInTerminal` for debuggee when --external-interpreter.


Saizan:
Currently, if the client supports `runInTerminal` we ask it to run `hdb proxy` which is just a proxy for the stdin/stdout/stderr of the debugee.
Let's instead ask the client to run the debugee directly, so no proxying of std* is needed. With --external-interpreter the debugee is its own dedicated process anyway.

Complications:
- We still need `hdb proxy` as it is now for  --internal-interpreter, because --external-interpreter does not handle GHC plugins.
- If the client does not support `runInTerminal` then we still have to handle proxying the std* between the main process and the debugee.

alt-romes:
> If the client does not support runInTerminal then we still have to handle proxying the std* between the main process and the debugee.

I'm not sure, if the client does not support runInTerminal we will launch the external interpreter in the same way, it just won't be connected to a user terminal (so not possible to give input interactively)

Saizan:
@alt-romes what about when the client starts the hdb process itself, doesn't hdb then try to get input/output for the debugee from its own stdin/stdout? Does that work with --external-interpreter?

PR #260, opened 2026-04-02
Run external interpreter directly in terminal (#259) and Refactor DAP Haskell Testsuite

alt-romes:
See each commit individually.

- Does a first refactor to runInTerminal which enables running the external interpreter in the terminal directly
- Refactors the testsuite significantly to make testing more robust and extensible


alt-romes:
~~TODO We don't get the external interpreter PID directly, we just get the `shellProcessId`~~

alt-romes:
here's an idea though, I could exchange a message before starting the external interpreter which sends the PID over the socket

EDIT: that's what we did 👍

test/haskell/Test/DAP/Init.hs
```diff
@@ -31,28 +43,31 @@ startTestDAPServer :: FilePath -> [String] -> IO TestDAPServer
...
+              eof <- hIsEOF h
+              if eof
+                then return ()
+                else do
+                  l <- hGetLine h
+                  modifyIORef' serverOutputRef (l :)
+                  loop
+        loop
+
+  _ <- forkIO $ race_ (forwardServerHandle hout) (forwardServerHandle herr)
```
Saizan: I don't think we want to cancel one when the other returns.

alt-romes: Yeah, you're right. Indeed they have an exit condition which allows them to return without an exception

test/haskell/Test/DAP/Messages.hs
```diff
@@ -50,75 +45,46 @@ newtype TestDAP a = TestDAP { runTestDAP :: TestDAPClientContext -> IO a }
...
-responseMatch cmd =
-  MessageMatch ("response(" ++ cmd ++ ")") $ \v ->
-    maybe False id $ parseMaybe (withObject "response" $ \o -> do
-      typ <- o .: "type"
-      responseCmd <- o .: "command"
-      success <- o .: "success"
-      pure ((typ :: String) == "response" && (responseCmd :: String) == cmd && (success :: Bool))
-      ) v
+    withAsync (runTestDAP waitForResponse ctx) $ \v ->
+      runTestDAP (k v) ctx
```
Saizan, line 77: Shouldn't waitForResponse use the seqNum to make sure it's getting the right response?

alt-romes: It probably should, but in practice it's always in sequence.

I think that would be a good improvement in a follow up patch, but not for this one, if you agree

Saizan: I see, I guess it makes sense because we send responses only from the main connection thread

PR #264, opened 2026-04-08
CI: Cache GHC for macOS

Saizan:
The macOS runner doesn't come with ghc and we are reinstalling it from scratch every time, making the macOS setup step take 5 minutes rather than seconds.

We cache only on macOS just to limit cache size usage (apparently it's only 10GB total)


alt-romes:
Good idea 👍 thanks for spotting

alt-romes:
Land when ready :+1:

Issue #271, opened 2026-04-13
Parametrize `initDebugger` to take GHC session rather than use HieBios


Saizan:
Currently we give HieBios a project root and entry file and it gives us the GHC flags and units to load. These are ultimately passed to `runDebugger` to setup the session (i.e. HscEnv) for the debugger.

When integrated with the persistent worker hopefully we can get a ghc session directly from it? 

TODOs:
  - [ ] Scope out what API the persistent worker offers, maybe mock it as code in hdb codebase?
  - [ ] Define abstraction for "project session provider" that covers both HieBios and persistent worker.
  - [ ] Parametrize initDebugger/debuggerThread/runDebugger to use the new abstraction.
  - [ ] Implement project session provider abstraction with HieBios (default) , and mocked persistent worker API (if we have it).

Saizan:
Skipped the mocked API part. 

Fixed in b32ada0fd667cff023dac72b419cd25db280e62d.

PR #274, opened 2026-04-16
[#271] runHDBServer parametrized on DebugRunner

Saizan:
`runDebugger` now takes a `DebugRunner` parameter which takes care of providing a session and an action to load the debuggee home units.

The `DebugRunner` parametrization is pushed up the call stack until we can define a 
`runHDBServer` function which could be used from elsewhere to start a DAP server.


alt-romes:
The ubuntu runner was hanging.

I've made some changes to the error/request abortion cycle which should make the DAP testsuite better at exiting with an error rather than hanging in #260 (notably, 5a163db51d3efa24a49fd42cabee1c7b77ac7ce0)

Saizan:
@alt-romes about the Sdist tests, should I handle the source-package-repository dependence on `dap` like haskell-debugger-view or would it make sense to release dap-0.6 ?

alt-romes:
> @alt-romes about the Sdist tests, should I handle the source-package-repository dependence on `dap` like haskell-debugger-view or would it make sense to release dap-0.6 ?

@Saizan

Hmm. We need to make sure dap is released when we do a release, which is why this is a good CI check.

A few ideas:

1) Temporarily add a `source-repository` package stanza to a temporary cabal.project file in the sdist test
2) Make the sdist test only run on the release pipeline (but then subtle packaging bugs might be caught too late and it'd be annoying)
3) Release 0.6 now and more versions as necessary

The benefit of not releasing immediately is bundling more changes into a single release, but then again it's not that much of a worry to release a lot. Especially since we don't necessarily envision changing `dap` a lot.

@dmjio is releasing often fine for you? Thanks!

alt-romes:
I've opened https://github.com/haskell-debugger/dap/issues/33 tracking the new dap release

alt-romes:
(Please add a descriptive commit message. It will help when looking at the log.)

(It should mention `dap` is a library now for instance)

Reviewing now.

haskell-debugger.cabal
```diff
@@ -155,10 +155,9 @@ library
     hs-source-dirs:   haskell-debugger
     default-language: GHC2021
 
-executable hdb
+library dap
```
alt-romes, line 158: This change should be mentioned in the commit message

cabal.project
```diff
@@ -13,3 +13,7 @@ if !os(windows)
 if !os(windows)
     executable-dynamic: True
 
+source-repository-package
+  type: git
+  location: https://github.com/haskell-debugger/dap.git
+  tag: 2267060ddcee86a5d80b9d7ab29816deddcfe99c
```
alt-romes: Let's release `dap` 0.6. We'll drop this before landing.

hdb-dap/Development/Debug/Adapter/Init.hs
```diff
@@ -269,25 +281,22 @@ mkRunInTerminalThreads l RunProxyInTerminal{..} _
...
                -> MVar D.Command  -- ^ Read commands
                -> MVar D.Response -- ^ Write reponses
-               -> (DebugAdaptorCont () -> IO ())
-               -- ^ Allows unlifting DebugAdaptor actions to IO. See 'registerNewDebugSession'.
                -> IO ()
-debuggerThread l HieBiosFlags{..} extraGhcArgs mainFp runConf requests replies withAdaptor = do
-
+debuggerThread l debugRunner runConf requests replies = do
+{-
+-- FIXME
```
alt-romes: FIXME? Maybe this can be moved somewhere else?
Or the full debug command produced somewhere? It feels like this might be quite important to debug why the debugger might not successfully compile a given user program.

haskell-debugger/GHC/Debugger/Session.hs
```diff
@@ -221,21 +234,24 @@ setExposedInUnit unitId exposed = do
...
                   (unitIdString uid)
                   (UnitIdArg $ RealUnit (Definite uid))
                   (ModRenaming True [])
-              | uid <- exposed
+              | uid <- exposed ++ Set.toList home_units
               , uid /= rtsUnitId
               , uid /= ghcInternalUnitId
-              , unitIdString uid /= "haskell-debugger-view-in-memory"
+--              , unitIdString uid /= "haskell-debugger-view-in-memory"
+              , uid /= unitId
```
alt-romes, line 246: Do you recall what motivated this change? I don't follow immediately. Mostly curious.

haskell-debugger/GHC/Debugger/Session.hs
```diff
@@ -221,21 +234,24 @@ setExposedInUnit unitId exposed = do
...
         Just hue -> hue
         Nothing -> error $ "setExposedInUnit: unit not found " ++ unitIdString unitId
+  let home_units = allUnits $ hsc_HUG env
 
   let dflags = (homeUnitEnv_dflags old_ie) { packageFlags = [ExposePackage
                   (unitIdString uid)
                   (UnitIdArg $ RealUnit (Definite uid))
                   (ModRenaming True [])
-              | uid <- exposed
+              | uid <- exposed ++ Set.toList home_units
```
alt-romes: What does this mean? Is this part of the refactor or a more fundamental change? Perhaps one that should be done in a separate commit? Perhaps the ones below too? (with a proper independent explanation)

hdb/Main.hs
```diff
@@ -83,24 +68,31 @@ main = do
     HdbDAPServer{port, internalInterpreter, disableIpeBacktraces} -> do
       setBacktraceMechanismState IPEBacktrace (not disableIpeBacktraces)
       config <- getConfig port
+      hdbProgram <- getExecutablePath
```
alt-romes, line 72: add a comment here (perhaps copy from the location where this was deleted): the same program invoked with `external-interpreter` serves as the external interpreter

hdb/Main.hs
```diff
@@ -83,24 +68,31 @@ main = do
...
-          (talk l init_var internalInterpreter)
+          (talk l servConf internalInterpreter)
           (ack l )
     HdbCLI{..} -> do
         setBacktraceMechanismState IPEBacktrace (not disableIpeBacktraces)
         l <- mainLogger hdbOpts.verbosity stdout
         stdinStream <- case debuggeeStdin of
           Just fp -> UseHandle <$> System.IO.openFile fp ReadMode
           Nothing -> pure Inherit
+        thisProg <- getExecutablePath
```
alt-romes, line 91: same comment

haskell-debugger/GHC/Debugger/Monad.hs
```diff
@@ -203,284 +204,326 @@ data RunDebuggerSettings = RunDebuggerSettings
         --
         -- Left: we launch our own external interpreter process through
         -- GHC's spawnIServ using the given StdStream as the stdin.
+      , externalInterpreterProg :: FilePath
```
alt-romes, line 207:
```suggestion
      , externalInterpreterProg :: FilePath
        -- ^ The full path to the program we'll run as the external interpreter
```

haskell-debugger/GHC/Debugger/Monad.hs
```diff
@@ -203,284 +204,326 @@ data RunDebuggerSettings = RunDebuggerSettings
...
+        `GHC.gopt_set` GHC.Opt_IgnoreOptimChanges
+        `GHC.gopt_set` GHC.Opt_IgnoreHpcChanges
+        `GHC.gopt_set` GHC.Opt_UseBytecodeRatherThanObjects
+        `GHC.gopt_set` GHC.Opt_InsertBreakpoints
+
+        -- Enable the external interpreter by default! See #169
+        -- See Note [Custom external interpreter]
+        & enableExternalInterpreter conf.preferInternalInterpreter
+        -- Ext interp is the same program as this, with "--external-interpreter"
+        -- (this is ignored on GHC 9.14, see Note [Custom external interpreter])
```
alt-romes, line 304:
```suggestion
        -- Ext interp is typically the same program as this, with "--external-interpreter"
        -- (See Note [Custom external interpreter])
```

alt-romes (commented):
Looks good to me!

haskell-debugger/GHC/Debugger/Session.hs
Saizan: I guess this is more related to the other PR I opened, I'll move these changes to setExposedInUnit there if the tests still pass.

The `++ Set.toList home_units` change should be made where we calculate `exposed` anyway.

To explain though: even home units need to be exposed with an ExposePackage flag if we want IIDecl (package?) imports to include them when resolving the module.

haskell-debugger/GHC/Debugger/Session.hs
Saizan: as above, you do need `haskell-debugger-view-in-memory` to be exposed. I believe earlier we were getting errors about it because we were not including it in the home_units arg of initUnits

alt-romes:
@Saizan I've done the 0.6 `dap` release now: https://github.com/haskell-debugger/dap/issues/33. Update this patch accordingly (drop the cabal.project thing)

Saizan:
@alt-romes done, good to merge?

alt-romes:
Ye, land it!

alt-romes (approved):

PR #275, opened 2026-04-20
[Fix #242] Bring Interactive Imports in scope after resumeExec

Saizan:
Turns out the culprit for #242 was resumeExec resetting the scope to the one before the last break, with no care for consistency with the `ic_imports` field.

That also means we can just setContext again with the saved imports to get our scope right.

### DIFF
--- a/haskell-debugger/GHC/Debugger/Session.hs
+++ b/haskell-debugger/GHC/Debugger/Session.hs
@@ -27,6 +27,7 @@ module GHC.Debugger.Session (
   setPgmI, addOptI,
   setDynFlagWays,
   makeDynFlagsAbsoluteOverall,
+  resumeExec,
   )
   where
 
@@ -69,6 +70,7 @@ import qualified GHC.Unit.Home.Graph as HUG
 import qualified Data.Set as Set
 import Data.Maybe
 import GHC.Types.Target (InputFileBuffer)
+import GHC (SingleStep, ExecResult)
 
 -- | Throws if package flags are unsatisfiable
 parseHomeUnitArguments :: GhcMonad m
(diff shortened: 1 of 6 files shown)


Saizan:
I think this could probably be cherry-picked onto master without conflicts if it makes more sense

haskell-debugger/GHC/Debugger/Session.hs
```diff
@@ -482,6 +484,29 @@ addWay' w dflags0 =
...
    in dflags3
 
+-- ----------------------------------------------------------------------------
+-- Wrappers around GHC's odd behavior
+-- ----------------------------------------------------------------------------
+
+resumeExec :: GhcMonad m => SingleStep -> Maybe Int -> m ExecResult
+resumeExec a b = do
+  v <- GHC.resumeExec a b
+  -- resumeExec resets the scope of the IC to what it was before the last
```
alt-romes: It's unclear if you're referring to GHC's behavior or our new overwritten behavior

alt-romes: In general, this comment is a bit confusing because this distinction isn't clear

alt-romes (commented):
Great finding!! I think the wording could just be improved because I couldn't follow it very clearly

alt-romes:
Could you make this MR against master? I agree

haskell-debugger/GHC/Debugger/Session.hs
Saizan: @alt-romes should be better now

alt-romes (approved):
Much better! Could you rebase on master and discard all commits except the top-most? I can help you if it's not clear how to do that.

Basically, make the MR just to merge that last commit to master

PR #276, opened 2026-04-21
Add more regression tests

alt-romes:
Cherry-picks #253 #252 #250

Supersedes them


alt-romes:
Nevermind, let's keep going with the individual MRs because this still triggers T130b

PR #279, opened 2026-04-22
Refactor home unit session initialisation

fendor:
There were a couple of inaccuracies in the current session initialisation options.
We refactor a bit and make it easier to grasp what is going on.

### DIFF
--- a/haskell-debugger/GHC/Debugger/Session/Builtin.hs
+++ b/haskell-debugger/GHC/Debugger/Session/Builtin.hs
@@ -100,7 +100,7 @@ addInMemoryHsDebuggerViewUnit base_uids initialDynFlags = do
         , importPaths = []
         , packageFlags =
           [ ExposePackage
-                  ("-package-id " ++ unitIdString unitId)
+                  (unitIdString unitId)
                   (UnitIdArg $ RealUnit (Definite unitId))
                   (ModRenaming True [])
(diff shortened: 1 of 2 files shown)


haskell-debugger/GHC/Debugger/Session/Builtin.hs
```diff
@@ -100,7 +100,7 @@ addInMemoryHsDebuggerViewUnit base_uids initialDynFlags = do
         , importPaths = []
         , packageFlags =
           [ ExposePackage
-                  ("-package-id " ++ unitIdString unitId)
```
alt-romes, line 103: Really? how did it even work before?

alt-romes (commented):
LGTM

haskell-debugger/GHC/Debugger/Session/Builtin.hs
fendor: No idea how this ever worked

Saizan: This string doesn't matter, does it? It's only there to remember if the user wrote -package or -package-id.

The real argument is the UnitIdArg part

alt-romes: Makes sense, good point

Saizan (approved):

PR #280, opened 2026-04-22
Migrate NodeJS testsuite to Haskell driver

### LINKED ISSUE #269: Migrate NodeJS integration testsuite to new Haskell DAP testsuite
Since #260, the haskell testsuite for DAP is more robust and useful and extensible.

We should migrate the entire integration testsuite which uses the vscode-nodejs-test-debug-adapter library to test the DAP server to instead use our Haskell testsuite driver for DAP tests. It will be easier to improve it and fix in the future.

It's also better to do this now while the testsuite size is still manageable. We try to add all tests to the haskell testsuite already anyway, it's just a burden to maintain both testsuites.

alt-romes:
Fixes #269


alt-romes:
For CI we now do 2 jobs: one where we run the tests on the git checkout (for all platforms), and 1 job which runs with -Werror on ubuntu the testsuite *from* the sdist directly (ie not in tree)

PR #281, opened 2026-04-23
Fix testcase T130b by Exposing packages from module graph

### LINKED ISSUE #224: Test T130b fails locally and is blocking CI
While T130b passes in CI, it fails when running the testsuite locally.

Must fix!

Saizan:


alt-romes:
This makes sense to me, but given how much time we've spent chasing this bug it would be good to offer a more comprehensive commentary and some -- !! WARNING !! comment or something about how important it is.
(that is, I think a `Note` is warranted in the location where you call "`setExposeUnit ... final_mod_graph`.

`Note [Why the final mod graph units must be exposed explicitly]` or something like that.

The commit message should reference the note and #224

alt-romes:
Does this also fix #238 ?

Saizan:
@alt-romes seems to fix #238 when not using a multi-repl (so the breakpoint needs to be in the hdb exe)

alt-romes:
> @alt-romes seems to fix #238 when not using a multi-repl (so the breakpoint needs to be in the hdb exe)

Interesting! Sounds good, let's do that and open a new ticket for what doesn't work exactly with multi-reps

Saizan:
@alt-romes pushed the note

Also changed it so the units we expose are the same ones used to find haskell-debugger-view

With that I only get problems for #238 when I run the (debugger) server from an empty directory. This compilation error shows up in the log:
```
{
    "body": {
        "output": "[22 of 52] Compiling GHC.Debugger.Session.Builtin ( /home/andrea/haskell-debugger/haskell-debugger/GHC/Debugger/Session/Builtin.hs, interpreted )[haskell-debugger-0.12.3.0-inplace]\n"
    },
    "event": "output",
    "type": "event"
}
<no location info>: error:
    haskell-debugger-view/src/GHC/Debugger/View/ByteString.hs: withBinaryFile: does not exist (No such file or directory)

```
which I guess it's fair enough, because the embedStringFile calls just use a relative path

haskell-debugger/GHC/Debugger/Monad.hs
```diff
@@ -421,6 +421,30 @@ runDebugger l rootDir compDir libdir units ghcInvocation' extraGhcArgs mainFp co
         when (GHC.failed success) $ liftIO $
           throwM DebuggerFailedToLoad
 
+        -- Note [Must explicitly expose module graph units]
```
alt-romes: You should move the Note below `Note [Dynamic dependencies for dynamic debugger]` on this module, and follow the style there (Note name, next line is `~~~~~~`).

The idea is the note is at the top level and may be referenced by more than one location and doesn't complicate the code by being written inline in the middle of it all.

See [Developing GHC > Comments and Notes](https://aosabook.org/en/v2/ghc.html)

haskell-debugger/GHC/Debugger/Monad.hs
```diff
@@ -421,6 +421,30 @@ runDebugger l rootDir compDir libdir units ghcInvocation' extraGhcArgs mainFp co
...
+        -- can import interactively (i.e. with GHC.setContext).
+        --
+        -- The `moduleNameProvidersMap` has so far only been required to expose,
+        -- with `ExposePackage` flags, the other home units. However since
+        -- exposing a package **does not** imply exposing its dependencies,
+        -- `mkUnitState` was free to make a choice for us to, e.g., expose
+        -- haskell-debugger-view-0.2.1.0-... and hide
+        -- haskell-debugger-view-0.2.0.0-..., while the latter is the one in the
+        -- graph. While we noticed with `hdv` this can happen with any dep of
+        -- the debuggee, causing problems at the prompt.
```
alt-romes: If you explicitly expose `haskell-debugger-view-0.2.0.0`, can `haskell-debugger-view-0.2.1.0` somehow be implicitly exposed too?

alt-romes: Or does the latter only get picked because there was no explicit exposed one?

alt-romes (commented):
Excellent @Saizan! Can you squash the two commits and add two lines to the commit description saying
```
Fix #224
Fix #238
```

alt-romes:
> which I guess it's fair enough, because the embedStringFile calls just use a relative path

That's right, I wouldn't expect that to work because of `embedStringFile`.

Excellent that #238 is fixed by this too!

alt-romes (approved):

haskell-debugger/GHC/Debugger/Monad.hs
Saizan: AFAIU within a unit exposing one hides the other, that's how we were getting the "hidden haskell-debugger-view-0.2.0.0" errors

Saizan: With this patch we are getting a moduleNameProvidersMap like this

```
GHC.Debugger.View.ByteString -> [haskell-debugger-view-0.2.1.0-07408397037260571bbb43df8171563563272313fc0e1d44ecbe94247d64898c:GHC.Debugger.View.ByteString -> hidden package,
                                  haskell-debugger-view-0.1.0.0-cb5cc57a3201894d30a97d95a1f27b60067ae17b4d6adbcb1d5efb3bcd8dee70:GHC.Debugger.View.ByteString -> hidden package,
                                  haskell-debugger-view-0.2.1.0-485ee981ab310332f3c47b6c6152d45b363aaa196c64e88e618138a303829605:GHC.Debugger.View.ByteString -> hidden package,
                                  haskell-debugger-view-0.1.0.0-0ebb86c4670f51466882b864d00204afbac54b16469b47f05764ca5eb4ca769a:GHC.Debugger.View.ByteString -> hidden package,
                                  haskell-debugger-view-0.2.0.0-52ddb3bad2a4102ae4edfefad3d6fe5050b8f4175993b0a200186c71f37b526a:GHC.Debugger.View.ByteString -> exposed package,
                                  haskell-debugger-view-0.2.0.0-279c347bd3a91676a96135a63cdb9fc7f289a6ccc10db93f3b1e9b53930ca231:GHC.Debugger.View.ByteString -> hidden package],
 GHC.Debugger.View.Class -> [haskell-debugger-view-0.2.1.0-07408397037260571bbb43df8171563563272313fc0e1d44ecbe94247d64898c:GHC.Debugger.View.Class -> hidden package,
                             haskell-debugger-view-0.1.0.0-cb5cc57a3201894d30a97d95a1f27b60067ae17b4d6adbcb1d5efb3bcd8dee70:GHC.Debugger.View.Class -> hidden package,
                             haskell-debugger-view-0.2.1.0-485ee981ab310332f3c47b6c6152d45b363aaa196c64e88e618138a303829605:GHC.Debugger.View.Class -> hidden package,
                             haskell-debugger-view-0.1.0.0-0ebb86c4670f51466882b864d00204afbac54b16469b47f05764ca5eb4ca769a:GHC.Debugger.View.Class -> hidden package,
                             haskell-debugger-view-0.2.0.0-52ddb3bad2a4102ae4edfefad3d6fe5050b8f4175993b0a200186c71f37b526a:GHC.Debugger.View.Class -> exposed package,
                             haskell-debugger-view-0.2.0.0-279c347bd3a91676a96135a63cdb9fc7f289a6ccc10db93f3b1e9b53930ca231:GHC.Debugger.View.Class -> hidden package],
 GHC.Debugger.View.Containers -> [haskell-debugger-view-0.2.1.0-07408397037260571bbb43df8171563563272313fc0e1d44ecbe94247d64898c:GHC.Debugger.View.Containers -> hidden package,
                                  haskell-debugger-view-0.1.0.0-cb5cc57a3201894d30a97d95a1f27b60067ae17b4d6adbcb1d5efb3bcd8dee70:GHC.Debugger.View.Containers -> hidden package,
                                  haskell-debugger-view-0.2.1.0-485ee981ab310332f3c47b6c6152d45b363aaa196c64e88e618138a303829605:GHC.Debugger.View.Containers -> hidden package,
                                  haskell-debugger-view-0.1.0.0-0ebb86c4670f51466882b864d00204afbac54b16469b47f05764ca5eb4ca769a:GHC.Debugger.View.Containers -> hidden package,
                                  haskell-debugger-view-0.2.0.0-52ddb3bad2a4102ae4edfefad3d6fe5050b8f4175993b0a200186c71f37b526a:GHC.Debugger.View.Containers -> exposed package,
                                  haskell-debugger-view-0.2.0.0-279c347bd3a91676a96135a63cdb9fc7f289a6ccc10db93f3b1e9b53930ca231:GHC.Debugger.View.Containers -> hidden package],
 GHC.Debugger.View.Text -> [haskell-debugger-view-0.2.1.0-07408397037260571bbb43df8171563563272313fc0e1d44ecbe94247d64898c:GHC.Debugger.View.Text -> hidden package,
                            haskell-debugger-view-0.1.0.0-cb5cc57a3201894d30a97d95a1f27b60067ae17b4d6adbcb1d5efb3bcd8dee70:GHC.Debugger.View.Text -> hidden package,
                            haskell-debugger-view-0.2.1.0-485ee981ab310332f3c47b6c6152d45b363aaa196c64e88e618138a303829605:GHC.Debugger.View.Text -> hidden package,
                            haskell-debugger-view-0.1.0.0-0ebb86c4670f51466882b864d00204afbac54b16469b47f05764ca5eb4ca769a:GHC.Debugger.View.Text -> hidden package,
                            haskell-debugger-view-0.2.0.0-52ddb3bad2a4102ae4edfefad3d6fe5050b8f4175993b0a200186c71f37b526a:GHC.Debugger.View.Text -> exposed package,
                            haskell-debugger-view-0.2.0.0-279c347bd3a91676a96135a63cdb9fc7f289a6ccc10db93f3b1e9b53930ca231:GHC.Debugger.View.Text -> hidden package],
```

alt-romes: Ah! Makes sense.

Land when ready.

alt-romes: > With this patch we are getting a moduleNameProvidersMap like this

This is perfect, just the correct versions are exposed 👍

PR #284, opened 2026-04-23
Fix hd-view package-qualified imports bypassing home-unit

### LINKED ISSUE #283: Package-qualified imports bypass home units
@Saizan figured it out:

- Package-qualified imports look directly for exposed packages, IGNORING the home units modules.
- If we have INSTALLED `package-A-0.1`, and LOAD into hdb/ghci `package-A-0.2`, both which implement `VersionA.myVersion::IO()`:
    - Calling `myVersion` will print `0.2`.
    - However, `import "package-A" VersionA` *then* `myVersion` will print 0.1

The fact that package-qualified imports bypass the home unit modules may result in these very confusing scenarios where we have two different versions of the same package "available" when package-qualified-imports are used.

This resulted in a really weird bug when trying to run the debugger in itself (see also #245), where `haskell-debugger-view` is a home-unit (it's listed in the cabal.project), but when the debugger is initialized we try to load `haskell-debugger-view` using a package-qualified import. This would fail because the package referenced wasn't installed!

What does this mean?

- For hdb itself, if the `haskell-debugger-view` unit is part of the home-units, then we should NOT package-qualify it.
- For users, I think there's little we can do:
    - In a ghci session, trying to package-qualify import a package to refer to the package which is itself loaded into the session doesn't make a lot of sense. The package is being interpreted live, it's not installed anywhere so it can't be referenced.
    - If a user accidentally package-imports its own package, well, hope for the best? Hopefully they'll find this ticket!

alt-romes:
See Note [Do not package-qualify imports for home units]

Fixes #283

### DIFF
--- a/haskell-debugger/GHC/Debugger/Monad.hs
+++ b/haskell-debugger/GHC/Debugger/Monad.hs
@@ -94,6 +94,8 @@ import GHC.Platform.Ways
 import GHC.Data.FastString.Env (emptyFsEnv)
 #endif
 
+import GHC.Unit.Home.Graph
+
 -- | A debugger action.
 newtype Debugger a = Debugger { unDebugger :: ReaderT DebuggerState GHC.Ghc a }
   deriving ( Functor, Applicative, Monad, MonadIO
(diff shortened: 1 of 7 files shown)


haskell-debugger/GHC/Debugger/Monad.hs
```diff
@@ -533,6 +535,33 @@ An alternative, closer to what ghci does, would be to copy the `packageFlags`
...
+A package-qualified module import will be looked up directly in the exposed
+packages, IGNORING the home units modules.
+
+This can lead to two scenarios:
+
+  1) a package-import of a loaded unit fails, because that unit, despite being
+  loaded, is not installed
+
+  2) a package-import of a loaded unit succeeds, because a unit with the same
+  name (but not necessarily the same unit-id!), is installed.
```
Saizan, line 549: Maybe clarify that you mean `import "haskell-debugger-view" ...` can be ambiguous, at it's not a full unitId. But we use the full unitIdString in the `RawPkgQual` which sould be unambiguous?

Saizan (approved):

haskell-debugger/GHC/Debugger/Monad.hs
alt-romes: That's us, but a user can use a package-qualified import *without* it being the specific unit-id.

Sorry, I missed this comment before merging.

Issue #301, opened 2026-04-29
How to persist DebugView thunks which are not part of the input value


alt-romes:
Repro `Main.hs`:
```hs
module Main where
import GHC.Debugger.View.Class

data X = X String
    deriving Show

instance DebugView X where
    debugValue _ = simpleValue "T300-X" True
    debugFields (X s) = pure $ VarFields
        [ ("thunkInt", VarFieldValue intThunk)
        ]
      where
        intThunk :: Int
        intThunk = sum [1 .. length s + 99]

main :: IO ()
main = f (X "A")

f :: Show a => a -> IO ()
f action = do
    print action
```

`T300.cabal`:
```
cabal-version:   3.14
name:            T300
version:         0.1.0.0
license:         NONE
author:          Rodrigo Mesquita
maintainer:      rodrigo.m.mesquita@gmail.com
build-type:      Simple

common warnings
    ghc-options: -Wall

executable t300
    import:           warnings
    main-is:          Main.hs
    build-depends:    base, containers
    build-depends:    haskell-debugger-view
    hs-source-dirs:   .
    default-language: Haskell2010
```

The problem happens if you break on `print action`, look at `action` in the variables pane (which uses `DebugView`), force the `thunkInt` field of the `action`, and observe that it displays the right value momentarily, but immediately is cleared to a thunk again.

The "clear" is caused by the invalidate request which is sent whenever a value is forced, triggering a re-draw of the variables.
But re-requesting the variable of the `DebugView` field should now display the thunk, because the variable has already been evaluated.

I'm going to push a test.

alt-romes:
Ahhhhh........... I guess this is to be expected when the `DebugView` field value is not an existing value which is persisted across the calls to `debugFields`, but is rather just an arbitrary computation.

That also explains why this works fine for IntMaps with lazy values.

Every time we recompute the debug view fields we re-create from scratch that thunk, and it must be forced every time from scratch.

alt-romes:
Would be interesting to think of a way to fix this.

The real-world example in which this came up was:

```hs
instance DebugView ModuleGraph where
  debugValue   _ = simpleValue "ModuleGraph" True
  debugFields mg = pure (VarFields [("$html", VarFieldValue (renderModuleGraphHtml mg))])
```

where `renderModuleGraphHtml :: ModuleGraph -> String`.

The problem is that this `$html` field, when forced, is immediately refreshed and cleared because we re-create the `renderModuleGraphHtml` computation from scratch.

[...]

PR #302, opened 2026-04-29
Draft: test: Add test for #301?

alt-romes:
If we use an arbitrary computation as the value for a DebugView field,
which is re-created from scratch everytime we recompute that DebugView
fields, then evaluating that thunk is not persisted across refreshing
the variable.

Every time we re-fresh the variable we re-create the thunk, so it makes
some sense that forcing it is not persisted across re-freshing, but
still a bit unsatisfactory (esp. since the variables are invalidated as
soon as a variable is forced, so we only see a flash of these values)

See #301

### DIFF
--- a/test/haskell/Test/Integration/Variables.hs
+++ b/test/haskell/Test/Integration/Variables.hs
@@ -8,9 +8,7 @@ import qualified Data.Text as T
 import Test.DAP
 import Test.Tasty
 import Test.Tasty.HUnit
-#ifdef mingw32_HOST_OS
 import Test.Tasty.ExpectedFailure
-#endif
 import DAP (Variable, variableValue)
 
 variableTests :: TestTree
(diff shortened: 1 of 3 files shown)


alt-romes:
Maybe this shouldn't be considered unexpected behavior.

Would the user ever want these fields to be lazy? I guess if we don't want to do the work straight away but still have the option to display on force would be useful.

However, that means we'd need a way of persisting the value across invocations. Potentially by extending `Program` to allow explicitly allocating a re-usable value at a given key.

See https://github.com/well-typed/haskell-debugger/issues/301#issuecomment-4345121287

alt-romes:
I now think we should just accept that a `VarFieldValue` of a value always computed from scratch must be forced to display properly, otherwise it's just going to do the uniform thing which is display the right hand side as a thunk. And every time we load it again we see the thunk again as it's not preserved anywhere. That is the "natural" let's uniformly observe the value that `VarFieldValue` wraps directly

See the comments on #301

PR #293, opened 2026-04-24
Add test described by #233

alt-romes:

Issue #159, opened 2026-01-08
Add support for decoding custom stack annotations


alt-romes:
With the advent of callstacks (#107) we should consider adding support for decoding custom stack annotations (as per https://gitlab.haskell.org/ghc/ghc/-/issues/26218).

This would allow the callstack display pane to include the custom information put on the stack by the user.

@fendor might be interested in taking this one, since he originally implemented the custom stack annotations 😄

PR #306, opened 2026-05-04
[Fix #299] Be more specific about what hdb brings in scope.

### LINKED ISSUE #299: `Map` is still available at a breakpoint where `Map` should not be in scope
Consider:

```
module Main where

import Data.List (sort, nub)
import qualified Data.Map.Strict as Map
import Other (process)

main :: IO ()
main = do
  let xs = [3, 1, 4, 1, 5, 9, 2, 6] :: [Int]
  let m = Map.fromList [("a", 1), ("b", 2)] :: Map.Map String Int
  check xs m
  process xs

check :: [Int] -> Map.Map String Int -> IO ()
check xs m = do
  print xs
  print m
```

```
module Other where

process :: [Int] -> IO ()
process xs = do
  let n = length xs
  print n
```

Stop at the call to `process` in `Main.hs`, line 15. `Map` is in scope, so evaluating `Map.fromList [...]` at the REPL should succeed.

Then, run until you stop at a second breakpoint inside the definition of `process`, say line 4 (call to `print`). Evaluating `Map.fromList` at the REPL now shouldn't work (because `Map` is no longer in scope), but it does.

What do we make of that behavior? I'm not sure if that's expected.

Saizan:
Turns out if we are careful, and cheat about package names, we can actually import home units with IIDecl.

This makes it now possible to tailor the scope to be the one inside the Module of the breakpoint we are stopped at, as wished in #299.

- [x] IIModule import module of current breakpoint. 
- (Punted to issue #312) Extend debug session REPL with command to explicitly IIModule import some module?

### DIFF
--- a/haskell-debugger/GHC/Debugger/Monad.hs
+++ b/haskell-debugger/GHC/Debugger/Monad.hs
@@ -41,7 +41,6 @@ import System.Process.Internals (mkProcessHandle)
 import Text.Read (readMaybe)
 
 import GHC
-import GHC.Data.FastString
 import GHC.Data.StringBuffer
 import GHC.Driver.Config.Diagnostic
 import GHC.Driver.Config.Logger
@@ -61,9 +60,7 @@ import GHC.Runtime.Interpreter as GHCi
 import GHC.Runtime.Loader as GHC
 import GHC.Runtime.Context as GHCi
 import GHC.Types.Error
-import GHC.Types.PkgQual
 import GHC.Types.SourceError
-import GHC.Types.SourceText
 import GHC.Types.Unique.Supply as GHC
 import GHC.Unit.Module.Graph
 import GHC.Unit.State
@@ -399,8 +396,6 @@ runDebuggerAction l rootDir extraGhcArgs conf loadHomeUnit (Debugger action) = f
         GHC.initUniqSupply (GHC.initialUnique df) (GHC.uniqueIncrement df)
 
       loadHomeUnit
-      -- See Note [Must explicitly expose module graph units]
-      setExposedInUnit interactiveGhcDebuggerUnitId . graphUnits . hsc_mod_graph =<< getSession
 
       -- Ensure all the home units are built with same Ways and return them.
       buildWays       <- do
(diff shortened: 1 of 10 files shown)


alt-romes:
And we can only run REPL evaluation when we're stopped somewhere, so there's always a consistent answer for "what's in scope":

- The things we've imported manually
- The things imported by the module we're stopped at

Great!

alt-romes:
Should the test introduced #293 be marked as successful now?

Saizan:
@alt-romes I'm finding that we also need the scope of where we are stopped for `getStackTrace` otherwise T159 and `displays stack annotation frames (issue #159)` fail with `Not in scope: type constructor or class SomeStackAnnotation`. 
I guess that makes sense? Do you think there's other commands that might need it?

Saizan:
@alt-romes sorry, spoke too soon, I can trigger the getStackTrace problem on master with smaller imports: issue #310 , testcase #311 .

alt-romes:
Let me know once ready to review!

Saizan:
@alt-romes I believe it's ready now

haskell-debugger/GHC/Debugger/Monad.hs
```diff
@@ -413,35 +408,39 @@ runDebuggerAction l rootDir extraGhcArgs conf loadHomeUnit (Debugger action) = f
...
-                      , sl_tc = Nothing
-                      }
-                }) loadedBuiltinModNames
+            = map (packageImportDecl hvd_pkgName) loadedBuiltinModNames
+            where
+              hvd_pkgName = fromMaybe (error $ "No package name for: " ++ unitIdString hdv_uid) $
+                lookupUnitPackageName hsc_env_new hdv_uid
+
+      -- See Note [Must explicitly expose module graph units]
+      setExposedInUnit interactiveGhcDebuggerUnitId . graphUnits . hsc_mod_graph =<< getSession
```
alt-romes: Why do we call this twice?

haskell-debugger/GHC/Debugger/Monad.hs
```diff
@@ -413,35 +408,39 @@ runDebuggerAction l rootDir extraGhcArgs conf loadHomeUnit (Debugger action) = f
       -- in-memory sources.
       (hdv_uid, loadedBuiltinModNames) <- findOrLoadHaskellDebuggerView l buildWays
 
+      -- See Note [Must explicitly expose module graph units]
+      setExposedInUnit interactiveGhcDebuggerUnitId . graphUnits . hsc_mod_graph =<< getSession
```
alt-romes, line 412: The other call is above, here

haskell-debugger/GHC/Debugger/Monad.hs
```diff
@@ -413,35 +408,39 @@ runDebuggerAction l rootDir extraGhcArgs conf loadHomeUnit (Debugger action) = f
...
-          map (GHC.IIModule . GHC.ms_mod) mss)
+      let
+        imports
+          = preludeImp :
+            dbgViewImps ++
+            [ packageImportDecl pkgName (moduleName modl)
+            | modl <- map GHC.ms_mod mss
+            , let uid = moduleUnitId modl
+            , let pkgName = fromMaybe (error $ "No package name for: " ++ unitIdString uid) $ lookupUnitPackageName hsc_env_new uid
+            ]
```
alt-romes, line 443: Why do we import these home modules again? Perhaps you could add a short comment

haskell-debugger/GHC/Debugger/Monad.hs
```diff
@@ -413,35 +408,39 @@ runDebuggerAction l rootDir extraGhcArgs conf loadHomeUnit (Debugger action) = f
...
       let preludeImp = GHC.IIDecl . GHC.simpleImportDecl $ GHC.mkModuleName "Prelude"
+
+      hsc_env_new <- getSession
+
       -- dbgView should always be available, either because we manually loaded it
       -- or because it's in the transitive closure.
-      hug <- hsc_HUG <$> getSession
       let dbgViewImps
-            -- If hs-dbg-view is a home-unit, refer to it directly
-            -- See Note [Do not package-qualify imports for home units]
```
alt-romes, line 424: There used to be a reference to this note here. You've updated the note name, but it doesn't seem to be referenced from anywhere? Maybe here and below where we also use `packageImportDecl` it would be good to reference it

haskell-debugger/GHC/Debugger/Session.hs
```diff
@@ -363,6 +416,33 @@ mkSimpleTarget df fp = GHC.Target (GHC.TargetFile fp Nothing) True (homeUnitId_
...
+newtype PackageQualifier = PackageQualifier FastString
+
+lookupUnitPackageName :: HscEnv -> UnitId -> Maybe PackageQualifier
+lookupUnitPackageName env uid = home_unit_name <|> ext_unit_name
+          where
+            -- See Note [Package Qualified Imports]
+            home_unit_name = PackageQualifier . mkFastString <$> (thisPackageName =<< home_unit_dflags env uid)
+            ext_unit_name = const (PackageQualifier (unitIdFS uid)) <$> lookupUnitId (hsc_units env) uid
+
+packageImportDecl :: PackageQualifier -> ModuleName -> GHC.InteractiveImport
```
alt-romes: Referencing the note here sounds fine, but why is it referenced? I can't easily make the connection between the source code and what is written there.
A hint would be useful "do ... bc..., see Note ..."

test/golden/self-debug-cli/self-debug-cli.hdb-stdin
```diff
@@ -1,5 +1,6 @@
 break --name runDebugger
 run cli test/golden/self-debug-cli/Main.hs
+print import GHC.Data.FastString
```
alt-romes: This is surprising! If we're stopped at `runDebugger`, why do we need to implicitly export GHC.Data.FastString?

Shouldn't all imported module bindings be available at the breakpoint? That would be the most natural design. Having to re-import all modules to work with the data at a given breakpoint sounds not great

alt-romes (commented):
> only fully qualified names are accepted.

Hmm. Well, that's a tradeoff. I think it's fair, especially if everything is more natural (ie it's obvious what the context is - the breakpoint context - rather than a complicated magic context with implicit things)

> At the start of a session we import the **exported** symbols of:
> - Prelude
> - loaded Debug View modules
> - debugeee home units module

what do you mean by the third item in this list? I thought you changed it so we only have Prelude and the DebugView instances.

Could you also say in the commit message which ticket the commit fixes, if any? Here it's #299. This is important when I'm writing the changelog.

Overall: Looks good, but I have a few questions. I'm glad you understand all of these tricky bits now!

test/golden/self-debug-cli/self-debug-cli.hdb-stdin
Saizan: GHC.Data.FastString would be a redundant import from GHC.Debugger.Monad with these changes, so I moved the import here.

The `evaluateImportedBindings` already tests that we do have imported modules from the breakpoint in scope.

alt-romes: Ah! makes sense.

alt-romes (approved):

alt-romes:
Land it! Great work.

PR #308, opened 2026-05-04
HDB+GHC session for HLS and debug

Saizan:
See DEBUG.md for required setup and limitations

### DIFF
--- a/DEBUG.md
+++ b/DEBUG.md
@@ -1,0 +1,54 @@
+# Debugging hdb+GHC
+
+## prerequisites
+
+- GHC source tree at ./ghc with a ghc/compiler that allows Cabal-16.*
+  - For 9.14 cherry-pick b5508f2cb4e46aedb8f6232c1d42bd697e3dd2f0 
+    - See https://gitlab.haskell.org/trac-Saizan/ghc/-/commits/debug-9.14
+  - Gone through early stages of prep and build so `./ghc/_build/stageBoot/bin/` is populated. 
+- Installed ghc of same version and flavour as GHC source tree being debugged.
+- Installed hdb with executable-dynamic: True.
+
+## How To
+- Uncomment the bios cradle in the hdb's hie.yaml and comment the others.
+  - The bios cradle can also be left in place to have an HLS session with both ghc/compiler and hdb loaded.
+- Setup the final debuggee with a cradle that won't use package dbs setup by HIE.
+  - See Limitations
+  - See ./scripts/cabal-repl-bios.sh for a script that can be used as a bios cradle for this, uses cabal repl.
+- See .vscode/launch.json for how to start a Debug session of hdb.
+
+### Limitations
+
+#### Segfaults
+We are getting segfaults in the debugger hdb process, and this happens before running the debuggee. It's been useful to set `verbosity = 3` in `runDebugger` to have a rough idea of what can trigger the segfault.
+
+Current workarounds are:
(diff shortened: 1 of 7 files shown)


alt-romes:
I should also land my changes that introduce the view as HTML DebugView instance for `ModuleGraph` :)

DEBUG.md
```diff
@@ -0,0 +1,54 @@
+# Debugging hdb+GHC
+
+## prerequisites
+
+- GHC source tree at ./ghc with a ghc/compiler that allows Cabal-16.*
+  - For 9.14 cherry-pick b5508f2cb4e46aedb8f6232c1d42bd697e3dd2f0 
+    - See https://gitlab.haskell.org/trac-Saizan/ghc/-/commits/debug-9.14
+  - Gone through early stages of prep and build so `./ghc/_build/stageBoot/bin/` is populated. 
+- Installed ghc of same version and flavour as GHC source tree.
```
alt-romes:
```suggestion
- Installed ghc of same version and flavour as GHC source tree being debugged.
```

DEBUG.md
```diff
@@ -0,0 +1,54 @@
...
+- Installed hdb with executable-dynamic: True.
+
+## How To
+- Uncomment the bios cradle in hie.yaml and comment the others.
```
alt-romes:
```suggestion
- Uncomment the bios cradle in the hdb's hie.yaml and comment the others.
```

DEBUG.md
```diff
@@ -0,0 +1,54 @@
...
+  - This can also be left as is to have an HLS session with both ghc and hdb loaded.
```
alt-romes: what is left "as is"? `component: all`? Better to be explicit

DEBUG.md
```diff
@@ -0,0 +1,54 @@
...
+- Setup the final debuggee with a cradle that won't use package dbs setup by HIE.
+  - See Limitations
+  - See ./scripts/cabal-repl-bios.sh for a script that can be used as a bios cradle for this, uses cabal repl.
```
alt-romes, line 17: I don't understand this too well. I thought setting the hie.yaml to `debug-with-ghc-bios` to be sufficient? does this mean we have to do something else?

alt-romes (approved):
LGTM, just a few tweaks

DEBUG.md
Saizan: Yes, I was getting segfaults when the debugee's debugee was using a normal cradle and using the above script with a bios cradle helped.

So I suppose you can start without and look into this if you get a segfault while the units are getting setup. At some point we'll actually solve this.

DEBUG.md
```diff
@@ -0,0 +1,54 @@
+# Debugging hdb+GHC
+
+## prerequisites
+
+- GHC source tree at ./ghc with a ghc/compiler that allows Cabal-16.*
+  - For 9.14 cherry-pick b5508f2cb4e46aedb8f6232c1d42bd697e3dd2f0 
+    - See https://gitlab.haskell.org/trac-Saizan/ghc/-/commits/debug-9.14
+  - Gone through early stages of prep and build so `./ghc/_build/stageBoot/bin/` is populated. 
+- Installed ghc of same version and flavour as GHC source tree being debugged.
+- Installed hdb with executable-dynamic: True.
```
alt-romes, line 10: does this `hdb` have to be compiled by the same `ghc` as above? If yes say explicitly

alt-romes (approved):
lgtm (please squash into independent commits only)

Issue #310, opened 2026-05-06
getStackTrace should ensure required scope


Saizan:
The `getStackTrace` function can fail with

```
Not in scope: type constructor or class `SomeStackAnnotation'
```

if that symbol is not otherwise in scope (e.g. imported at current breakpoint).

alt-romes:
~~The comment in #306 mentions this is only needed for test T159  ? What's that mean?~~

fendor:
Should be fixed by #332

Saizan:
It is fixed by #332 i.e. 50d9a35648859b58e6f69a7d8ea47803dafbfba6

PR #311, opened 2026-05-06
test: add Test for #310

Saizan:
A variant of T159 with minimal imports. Recorded problem at issue #310.

TODO:
- [x] Mark it as expected failure


alt-romes:
was this meant to fail but it just succeeded? hah :;)

Saizan:
@alt-romes It succeeded as expected failure, I figured we'd want to set it up that way to merge the testcase first
``` 
T159b.fails-310:                                                              FAIL (expected: #310) (0.69s)
```

alt-romes:
Ahh, it's in the filename `fails-310`. OK.

So the recorded output is what you expected, but not what we get. Ok.

PR #313, opened 2026-05-08
Use dedicated hie-self-debug-test.yaml for self-debug (tests)

Saizan:
I kept getting annoyed that the tests would pick up configs from cabal.project.local so I added this extra config option:

- Added `cradleFile` Launch argument and --cradle-file option to support use of custom hie.yaml files for debug.
- self-debug tests specify cradleFile=hie-self-debug-test.yaml, which means:
  - using cradle with single "all" component
  - ignoring cabal.project.local

As a bonus we can now have the `multi` cradle in hie.yaml while using an "all" cradle for .vscode/launch.json:
- Added hie-self-debug.yaml which also uses "all" cradle but does not ignore cabal.project.local.
  - specified as cradleFile in .vscode/launch.json
- Uncommented multi cradles in default hie.yaml file, and removed "all" cradle.

We can get rid of hie-self-debug.yaml once we add the LoadAll/Eagerly load style to HIE.

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -9,7 +9,8 @@
             "entryFile": "hdb/Main.hs",
             "entryArgs": ["server", "--port", "12345"],
             "entryPoint": "main",
-            "extraGhcArgs": []
+            "extraGhcArgs": [],
+	    "cradleFile" : "hie-self-debug.yaml"
         },
         {
             "type": "haskell-debugger",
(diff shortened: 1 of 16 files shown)


alt-romes:
This is great! Means we can have HLS and the debugger both running without needing "multiComponents:" issues (not that I had them anymore but yeah).

alt-romes:
note: the self debug test can be debugged by calling `cabal sdist` and running the testsuite in the unzipped result

alt-romes (approved):

PR #318, opened 2026-05-20
expensiveLocals: TODO: Report bug with Locals scope, merge this as a …

alt-romes:
…workaround or when it really is too slow to expand by default


alt-romes:
The issue with not reporting issues promptly is that I no longer have any idea what this was referring to. I'm going to close it. If we come across anything like this again, it's not that good of a patch anyway so just do it from scratch and open appropriate tickets.

PR #321, opened 2026-05-28
Various changes needed for ghc-worker integration

Saizan:


alt-romes:
Let me know once I should look. Cool!

Saizan:
@alt-romes ready to review (prob. since before zurihac but forgot to ping)

haskell-debugger/GHC/Debugger/Session.hs
```diff
@@ -171,9 +174,8 @@ setupHomeUnitGraph flagsAndTargets = do
 -- | Set up the 'HomeUnitGraph' with empty 'HomeUnitEnv's.
 -- The first 'DynFlags' are the 'DynFlags' for the interactive session.
 createHomeUnitGraph :: GHC.Logger -> [DynFlags] -> IO HomeUnitGraph
-createHomeUnitGraph logger unitDflags0 = do
-  -- See Note [ Ambiguous Package Qualified Imports Workaround ]
-  let unitDflags = fixFlagsForIIDecl unitDflags0
+createHomeUnitGraph logger unitDflags = do
+  --let unitDflags = fixFlagsForIIDecl unitDflags0
```
alt-romes: should this be left as a comment? what does it mean? has the migration been done or not?

remove, or clarify

haskell-debugger/GHC/Debugger/Monad.hs
```diff
@@ -397,6 +400,9 @@ runDebuggerAction l rootDir extraGhcArgs conf loadHomeUnit (Debugger action) = f
 
       loadHomeUnit
 
+      fixHomeUnitsDynFlagsForIIDecl
+      GHC.modifySessionM $ liftIO . addInteractiveGhcDebuggerUnit
```
alt-romes: does this end up getting called twice? Once here, once in `initHomeUnitEnv`? Is this idempotent and/or expected?

haskell-debugger/GHC/Debugger/Breakpoint.hs
```diff
@@ -280,16 +280,16 @@ parseQC a (x:xs)       = parseQC (x:a) xs
 getModuleByPath :: FilePath -> Debugger (Either SDoc ModSummary)
 getModuleByPath path = do
   -- get all loaded modules this every time as the loaded modules may have changed
-  lms <- getAllLoadedModules
+  lms <- mapM (\ m -> fmap (m,) . liftIO . makeAbsolute . msHsFilePath $ m) =<< getAllLoadedModules
```
alt-romes: How often is `getModuleByPath` called? I reckon pretty often. Maybe `getAllLoadedModules` absolute paths should be somehow cache?

Feels like it would be slightly cleaner if `getAllLoadedModules` (INVARIANT: returns normalized Absolute paths) had a cache

alt-romes (approved):
Looks good to me, just a few comments

haskell-debugger/GHC/Debugger/Breakpoint.hs
Saizan: Not so often? Mostly when the dap-client queries or manipulates breakpoints.

Though now that I look at this again we should not `makeAbsolute` these paths, but rather use the workingDir from their units' DynFlags.

How would you invalidate a `loadedModules :: [ModSummary]` cache with patched paths though? Currently we only load modules in runDebugger I think, but I could see that changing.

PR #324, opened 2026-06-01
filter DapLibraryLog by verbosity too

Saizan:
`hdb server -v0 --port ...` should only output messages of severity Error, currently it prints more than that.

Here we fix the major source of messages: logging from the debug library.

Other sources:
- stray putStrLn
- GHC logger


alt-romes:
I don't think the GHC logger should be a problem, since all of its messages are logged via our logging mechanism (we hijack the logging mechanism to do this). (e.g. see `GHCLog logflags msg_class srcSpan msg ->` in `logDebugLog`)

The GHC logger messages like "Compiling ..." should be shown by default IMO.

alt-romes (approved):

PR #329, opened 2026-06-07
Remove the 'install using stack' section from the readme

alexj136:
Remove the 'install using stack' section from the readme as those instructions currently do not work.


alt-romes:
Thanks!

PR #255, opened 2026-03-31
Test external commands with nightly GHC

alt-romes:
Fixes #231

PR #332, opened 2026-06-19
Support GHC 10.1

fendor:


haskell-debugger/GHC/Debugger/Session/Interactive.hs
```diff
@@ -108,4 +108,4 @@ extendInteractiveContextWithIdsNoOccEnv ictxt new_ids
           }
     -- no change to ic_gre_cache as that deals with OccNames
   where
-    new_tythings = map AnId new_ids
+    new_tythings = map AnId new_ids
```
alt-romes:
```suggestion
    new_tythings = map AnId new_ids
```

alt-romes (approved):

alt-romes:
Perhaps you could try cherry-picking https://github.com/well-typed/haskell-debugger/pull/255 onto here too, to see if you can get 10.1 tested on CI

.github/workflows/debugger.yaml
```diff
@@ -115,25 +117,31 @@ jobs:
           (cd ${{ runner.temp }}/dist && tar xzf "haskell-debugger-view-${{env.haskellDebuggerViewVersion}}.tar.gz" && mv "haskell-debugger-view-${{env.haskellDebuggerViewVersion}}" "haskell-debugger-${{env.haskellDebuggerVersion}}/haskell-debugger-view")
           (cd ${{ runner.temp }}/dist/haskell-debugger-${{env.haskellDebuggerVersion}} && echo 'packages: . ./haskell-debugger-view' > cabal.project)
 
+      - name: Configure head.hackage for nightly
+        if: matrix.version == '10.1.20260422'
+        working-directory: ${{ runner.temp }}/dist/haskell-debugger-${{ env.haskellDebuggerVersion }}
+        run: |
+          curl --proto '=https' --tlsv1.2 -fsSL https://ghc.gitlab.haskell.org/head.hackage/cabal.project >> cabal.project.local
+          echo 'constraints: hashable == 1.4.7.0' >> cabal.project.local
+          cabal update
```
alt-romes: this might need tweaking or not be needed at all (you put hashable in cabal.project now iirc)

.github/workflows/debugger.yaml
```diff
@@ -70,16 +70,18 @@ jobs:
...
     continue-on-error: ${{ inputs.is-release == false }}
     strategy:
       matrix:
-        version: [9.14.1]
+        version: [9.14.1, 10.1.20260422]
         include:
           - channel: https://raw.githubusercontent.com/haskell/ghcup-metadata/refs/heads/develop/ghcup-prereleases-0.0.9.yaml
             version: 9.14.1
+          - channel: https://gitlab.haskell.org/ghc/ghcup-metadata/-/raw/updates/ghcup-nightlies-0.0.7.yaml
+            version: 10.1.20260422
```
alt-romes: update this to a more recent version^

alt-romes:
The 10.1 tests are hanging on ubuntu and macOS. Weird! Maybe something to do with signal handlers or termination or a deadlock? Very bizarre

alt-romes:
I canceled the tests on 10.1, they were hanging for the past 4h. How hard would it be to add a timeout to the GitHub Actions? Surely there's some existing mechanism. We should limit it to like 30m Max

alt-romes (approved):
Thanks, we can merge it like this. It's slightly unfortunate that many tests have duplicate test outputs (even if they are the same across versions), but no worries for now.

PR #335, opened 2026-06-22
[Fix #38] Use LoadUnitsFromCradle new hie-bios mode

### LINKED ISSUE #38: hie-bios only loads lazily components other than main by default
This is a bad default for the debugger (even though it is a good default for HLS?).
It's surprising that
```
cradle:
  cabal:
     - path: "./"
        component: "lib:hegg"
     - path: "./test/Test.hs"
        component: "test:hegg-test"
```
doesn't work, but if we instead force hie-bios to load all units eagerly:
```
cradle:
  cabal:
    component: "all"
```
it will.

Note that implicit-hie comes up with the former by default. So, by default, a user cannot set a breakpoint in a non-main unit.

- [ ] Duplicate test which sets breakpoint on internal unit but remove the hie.yaml file in it and make sure it still works.

Saizan:
Still draft because we should wait for new release of hie-bios.

TODO:
- [x] fix #ifdef to support older hie-bios versions
- [x] fix failing golden tests. Adding hie.yaml cradles if needed or just accepting the change in interpreted modules

### DIFF
--- a/hdb-dap/Development/Debug/Session/Setup.hs
+++ b/hdb-dap/Development/Debug/Session/Setup.hs
@@ -186,7 +186,9 @@ hieBiosFlags cradle root relTarget = runExceptT $ do
   -- because hie.yaml may invoke programs relative to the root (e.g. GHC's hie.yaml does)
   -- (HIE.getCompilerOptions depends on CWD being the proper root dir)
   let compilerOpts = liftIO $ withCurrentDirectory root $
-#if MIN_VERSION_hie_bios(0,14,0)
+#if MIN_VERSION_hie_bios(0,20,0)
+                          HIE.getCompilerOptions (HIE.TargetWithContext target [target]) HIE.LoadUnitsFromCradle cradle
+#elif MIN_VERSION_hie_bios(0,14,0)
                           HIE.getCompilerOptions target (HIE.LoadWithContext [target]) cradle
 #else
                           HIE.getCompilerOptions target [] cradle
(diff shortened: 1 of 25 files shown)


hdb-dap/Development/Debug/Session/Setup.hs
```diff
@@ -186,7 +186,9 @@ hieBiosFlags cradle root relTarget = runExceptT $ do
   -- because hie.yaml may invoke programs relative to the root (e.g. GHC's hie.yaml does)
   -- (HIE.getCompilerOptions depends on CWD being the proper root dir)
   let compilerOpts = liftIO $ withCurrentDirectory root $
-#if MIN_VERSION_hie_bios(0,14,0)
+#if MIN_VERSION_hie_bios(0,20,0)
+                          HIE.getCompilerOptions (HIE.TargetWithContext target [target]) HIE.LoadUnitsFromCradle cradle
```
fendor, line 189:
```suggestion
                          HIE.getCompilerOptions (HIE.TargetWithContext target []) HIE.LoadUnitsFromCradle cradle
```

This means that we always need a hie.yaml file, correct?

The trivial 
```yaml
cradle:
  cabal:
```

Doesn't work in the base case?

test/golden/T237/hie.yaml
```diff
@@ -0,0 +1,5 @@
+cradle:
+  cabal:
+    components:
+      - path: "app/"
+        component: "exe:T237"
```
fendor, line 5: This *shouldn't* be necessary, `haskell-debugger` ought to work in the simple case without any configuration.

Is this part of a test or strictly necessary?

Saizan: I wanted to match the session we'd have gotten with the old LoadWithContext style,  so that the semantics of the test didn't change, also having the type compiled and the instance interpreted seems the harder case to me.

Saizan: I added T237b which uses a `cradle: {"cabal": null }` hie.yaml, so I guess that resolves both issues?

Saizan:
@fendor @alt-romes is support for older hie-bios versions important? hie-bios itself supports all the way back to ghc ==9.6.7

alt-romes:
I think we only need to support the latest version if that simplifies the code and allows us to drop hacks.

fendor:
I see no benefit in supporting old hie-bios versions, the hie.yaml is backwards compatible and we don't drop support for any GHC version that haskell-debugger supports

test/golden/T237/hie.yaml
fendor: If that works as intended, yes!

Saizan:
Happy to merge this, then?

alt-romes:
Looks good to me, but what about this comment: https://github.com/well-typed/haskell-debugger/pull/335#discussion_r3512104292 ?

Saizan:
@alt-romes We have tests without hie.yaml and tests with hie.yaml with the barebone cradle above, so the new mode works in both cases.

LoadUnitsFromCradle fallbacks to the other modes if it can't find components in the cradle.

alt-romes (approved):

PR #338, opened 2026-07-08
[Fix #266] Add InterpreterSettings/DAPDebuggee interface

### LINKED ISSUE #266: Refactoring the runInTerminal vs external interpreter vs internal interpreter scenarios
There are 4 competing scenarios for (RunInTerminal, InternalInterpreter), (RunInTerminal, ExternalInterpreter), (NoRunInTerminal, InternalInterpreter), (NoRunInTerminal, ExternalInterpreter), where the synchronization and threads launched differ significantly.

We laid these out in the following sketches: (link: https://excalidraw.com/#room=0e4442275663c13a9943,mVf5TA6PS9pnBvzVX38HtQ)

Image

This ticket tracks refactoring the code to be more abstract wrt these 4 scenarios.

Saizan:
The second part of #266: the different ways of running a debuggee are now handled uniformly through an API that they all implement.

Everything to do with the `runInTerminal` feature is DAP specific, so we ended up with two abstractions rather than one:
- `InterpreterSettings` tells you what top-level DynFlags to set, and how to setup the GHC interpreter.
- `DAPDebuggee`: contains `InterpreterSettings` but also extra debug session threads, post-action (to send runInTerminal request), and a method to wait for the debuggee to be ready.

`hdb cli` only uses the `InterpreterSettings` abstraction, while the `hdb server` uses `DAPDebugee` (c.f. initDebugger).
 
We now throw an exception when we realize -f[no-]external-interpreter has been set in a way inconsistent with hdb's own flags.

Some code was found to be dead and removed.

Sockets opened to find an open port are kept open and used for the server, so we avoid a race condition on which session gets to use the port (though probably hard to trigger).


Saizan:
@alt-romes ready to review

hdb-dap/Development/Debug/Adapter.hs
```diff
@@ -74,6 +74,7 @@ data RunInTerminalProc
         -- When 'NoRunInTerminal' but using external-interpreter, we'll still
         -- launch the external interpreter but in the default GHC way using the
         -- file descriptors directly.
+      , socket :: Socket
```
alt-romes: what is this socket? add a haddock please

hdb-dap/Development/Debug/Adapter.hs
```diff
@@ -38,7 +38,7 @@ data DebugAdaptorState = DAS
       , entryPoint    :: String
       , entryArgs     :: [String]
       , projectRoot   :: AbsFilePath
-      , runInTerminalProc :: RunInTerminalProc
+      , waitForDebuggee :: IO ()
         -- ^ Potentially a process launched via 'runInTerminal'.
```
alt-romes: this comment is now outdated? what is `waitForDebuggee` in haddocks?

hdb-dap/Development/Debug/Adapter/Proxy.hs
```diff
@@ -135,10 +134,21 @@ labelMe name = do
...
   -- Bind on "0" to let the OS pick a free port
-  openTCPServerSocket addr
+
+  sock <- Network.Socket.socket (addrFamily addr) (addrSocketType addr) (addrProtocol addr)
+  -- Must set before bind!
+  setSocketOption sock ReuseAddr 1
+
+  bind sock (addrAddress addr)
+  listen sock maxListenQueue
+  return sock
```
alt-romes: semantic changes like these would ideally be done in a separate commit, but keep that in mind in the future. fine for now

alt-romes: why shouldn't we use openTCPServerSocket any longer ? not that I care about using `openTCPServerSocket` specifically, just curious about why the change

haskell-debugger/GHC/Debugger/Debuggee.hs
```diff
@@ -0,0 +1,267 @@
...
+                , cmdspec = case cmdspec cp of
+                    ShellCommand (words -> ws) -> ShellCommand $ unwords $ getPgmI dflags : drop 1 ws
+                    RawCommand _fp args -> RawCommand (getPgmI dflags) args
+#endif
+                }
+            putHandles (o, e)
+            return ph
+        }
+    }
+  let
```
alt-romes: what is this `let`? xd

haskell-debugger/GHC/Debugger/Debuggee.hs
```diff
@@ -0,0 +1,267 @@
...
+  -- we supply our custom external interpreter process, which is
+  -- already running and connected to the user's terminal.
+  extInterp <- liftIO
+    $ annotateStackStringIO "Waiting for an external interpreter run-in-terminal process"
+    $ m
+  modifySession $ \h -> h
+    { hsc_interp = Just extInterp -- set it directly!
+    }
+
+  -- Ext interp is running in user terminal, no need to forward output to logger
```
alt-romes: this comment doesn't make sense here; we don't know, in the GHC.Debugger side, whether the ext interp is running in the terminal or not. do we still have to care about forward logging with the new abstraction?

haskell-debugger/GHC/Debugger/Debuggee.hs
```diff
@@ -0,0 +1,267 @@
...
+  show (InconsistentInterpreterFlags t) = "Interpreter flags are inconsistent: " ++ t
+instance Exception InconsistentInterpreterFlags
+
+
+mkExternalInterpreterFromIOSetup :: IO Interp -> LogAction IO DebuggerLog -> DynFlags -> Ghc a -> Ghc a
+mkExternalInterpreterFromIOSetup m _l dflags mainGhcThread = do
+  unless (gopt Opt_ExternalInterpreter dflags) $ do
+    throw $ InconsistentInterpreterFlags $ "Used ghc flag  -fno-external-interpreter instead of --internal-interpreter haskell-debugger flag."
+  -- we supply our custom external interpreter process, which is
+  -- already running and connected to the user's terminal.
```
alt-romes: ditto, I don't think the abstraction should leak to GHC.Debugger side whether something is in terminal or not? not sure.

alt-romes (approved):
Excellent work. A few tweaks on little places but overall very good.

hdb-dap/Development/Debug/Adapter/Proxy.hs
Saizan: Good question, turns out we can, I'll make the change. 

My change was so we could integrate `setSocketOption sock ReuseAddr 1`  from one of the functions this one replaces, but `openTCPServerSocket` does that too, and the docs say so.

Saizan:
@alt-romes I was uneasy about removing those comments even if they didn't really belong. Anyway I took another look and we can actually:

- `Inherit` the stdout/stderr handles for `hdb cli`.
- Directly call Output.stdout/Output.stderr in the forwarding thread for the DAP server. That's enabled by registering the forwarding thread as part of the debug session, so we don't need the withAsync.

 Then we don't need `LogDebuggeeOut` and `LogDebuggeeErr` anymore.

Saizan:
@alt-romes turns out openTCPServerSocket causes problems on macOS, narrowed it down to this bit
```
      #if !defined(openbsd_HOST_OS)
        when (addrFamily addr == AF_INET6) $ setSocketOption sock IPv6Only 1
      #endif
```
Made a note of it in the code. We don't particularly care to force IPv6, let me know if you have some insight.

alt-romes:
> @alt-romes turns out openTCPServerSocket causes problems on macOS, narrowed it down to this bit
> 
> ```
>       #if !defined(openbsd_HOST_OS)
>         when (addrFamily addr == AF_INET6) $ setSocketOption sock IPv6Only 1
>       #endif
> ```
> 
> Made a note of it in the code. We don't particularly care to force IPv6, let me know if you have some insight.

LGTM, just keep what you had then.

PR #341, opened 2026-07-16
Call initUniqSupply only once.

Saizan:
I thought about setting up some way to still initialize on-demand but no more than once via an MVar, but then saw no real benefit since `initUniqSupply` doesn't actually construct anything to be passed around, it just pokes the values into the memory cells.

Settled for doing it in `main` then.

### DIFF
--- a/haskell-debugger/GHC/Debugger/Monad.hs
+++ b/haskell-debugger/GHC/Debugger/Monad.hs
@@ -54,7 +54,6 @@ import GHC.Runtime.Loader as GHC
 import GHC.Runtime.Context as GHCi
 import GHC.Types.Error
 import GHC.Types.SourceError
-import GHC.Types.Unique.Supply as GHC
 import GHC.Unit.Module.Graph
 import GHC.Unit.State
 import GHC.Unit.Module.ModSummary as GHC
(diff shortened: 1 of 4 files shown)


haskell-debugger/GHC/Debugger/Monad.hs
```diff
@@ -670,6 +666,30 @@ archive). This setting must definitely match the way in which the external
...
+
+The only safe time to initialize is if there are no existing Uniques that are still relevant,
+and since it's also cheap we do it right away in `main`, for the `cli` or `server` commands.
+
+Contrary to ghc itself, this means we do not honor the `initialUnique` and `uniqueIncrement`
+fields of DynFlags, but they seem to be there for testing anyway.
+
+Users of hdb-as-a-library, e.g. using runHDBServer, will have to do the initialization themselves,
+especially if they supply any loaded units/modules to us via the DebugRunner,
+as those will contain `Unique`s.
```
alt-romes, line 699: Maybe this should be added as a warning on the `runDebugger` haddocks, pointing to this note.

"INVARIANT: The initUniqSupply has already been initialized. (Add this to the function which used to call initUniqSupply)"

alt-romes (approved):
LGTM but address my comment please

PR #344, opened 2026-07-17
[Fix #273] Tag logs with session and thread Id + one more fix 

### LINKED ISSUE #273: Logs should reference the connection they belong to (if there's one).
best to use a legible pseudonym rather than an hash/uuid. We could set it as a ThreadLabel during initialize and then make it part of the log action when we initDebugger (and thread it wherever needed).

Saizan:
I had been making a few modifications while debugging the UniqueSupply bug, these should be kept.

- `DAPSessionLog` logs now have [SESSID=...][THREAD=<label or number>] tags.
- `DAPLog` logs only have the port, but one can relate it to the SESSID

- I was under the impression that `dap` would tear down the whole session on uncaught exceptions from any of the session threads. Instead it appears to only close the connection. No big deal though, we can have whichever crashed fork a thread to kill the rest.

### DIFF
--- a/haskell-debugger.cabal
+++ b/haskell-debugger.cabal
@@ -206,6 +206,7 @@ library dap-server
         optparse-applicative >= 0.18 && < 0.20,
         uuid >= 1.3 && < 1.4,
         ghc-stack-annotations >=0.1 && <0.2,
+        monad-control >= 1.0 && < 1.1,
 
 executable hdb
     import:           warnings
--- a/haskell-debugger/GHC/Debugger/Monad.hs
+++ b/haskell-debugger/GHC/Debugger/Monad.hs
@@ -33,6 +33,7 @@ import System.Posix.Signals
 #endif
 import qualified Data.List as L
 import qualified Data.List.NonEmpty as NonEmpty
+import qualified GHC.Conc.Sync as C
 
 import GHC
 import GHC.Data.StringBuffer
@@ -79,6 +80,8 @@ import GHC.Unit.Home.Graph
 import GHC.Debugger.Utils.Orphans () -- bring orphan instances to everything which uses `Debugger`
 import System.Directory (getCurrentDirectory)
 import GHC.Debugger.Debuggee
+import GHC.Plugins (HasCallStack)
+import Data.Bifunctor
 
 -- | A debugger action.
 newtype Debugger a = Debugger { unDebugger :: ReaderT DebuggerState GHC.Ghc a }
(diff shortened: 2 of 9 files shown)


haskell-debugger/GHC/Debugger/Monad.hs
```diff
@@ -395,6 +402,18 @@ runDebuggerAction l rootDir extraGhcArgs conf loadHomeUnit (Debugger action) = f
...
+  thId <- liftIO $ myThreadId
+  mlbl <- liftIO $ C.threadLabel thId
+  case mlbl of
+    Nothing -> m
+    Just lbl -> do
+      annotateCallStackGhc $ do
+        x <- m
+        liftIO $ C.labelThread thId lbl
+        pure x
+
```
alt-romes, line 416: That's funny looking! Why is this needed again? Fine by me

haskell-debugger/GHC/Debugger/Run.hs
```diff
@@ -126,6 +136,8 @@ debugExecution entryFile entry args = do
           { -- Disable dumping of any data during evaluation of GHCi's internal expressions. (#17500)
             dumpFlags = mempty
           }
+              `xopt_unset` LangExt.RebindableSyntax
```
alt-romes: hmm? if this is a fix it should really go in a separate commit so we can blame it later with a reason

why unset RebindableSyntax? I guess this is just for evaluating internal expressions?

haskell-debugger.cabal
```diff
@@ -206,6 +206,8 @@ library dap-server
         optparse-applicative >= 0.18 && < 0.20,
         uuid >= 1.3 && < 1.4,
         ghc-stack-annotations >=0.1 && <0.2,
+        uuid,
+        monad-control,
```
alt-romes: I don't think we are using this, even if the `dap` library does in some places. Are you sure we need it too?

hdb-dap/Development/Debug/Adapter.hs
```diff
@@ -49,6 +53,12 @@ data VariablesIx = VariablesIx StackFrameIx D.VariableReference
 instance MonadFail DebugAdaptor where
   fail a = sendError (ErrorMessage (T.pack a)) Nothing
 
+safeDestroyDebugSession :: Adaptor app request ()
+safeDestroyDebugSession = void $ do
+  -- Without forkIO we might kill ourselves first and not kill anything else.
+  liftBaseDiscard forkIO $
```
alt-romes, line 59: can't this just be `liftIO $ forkIO`?

alt-romes (commented):
Good work!

haskell-debugger/GHC/Debugger/Monad.hs
Saizan: `runPipelines` sets a "main --make thread" label

hdb-dap/Development/Debug/Adapter.hs
Saizan: This is where we are using `monad-control`: the call to `destroyDebugSession` happens in `Adaptor` because it needs to access the debug session state.

haskell-debugger/GHC/Debugger/Run.hs
Saizan: That's quite right, I missed it got in there when staging this commit.

I copied it from `GHCi.UI.runInternal` when debugging evalWrapper errors, seems we should actually copy the whole function and use it for all the "internal" calls like defineNoPrint etc.. I'll open an issue about it.

Saizan: https://github.com/well-typed/haskell-debugger/issues/347

Saizan:
@alt-romes should I feel free to merge PRs like this when I'm confident the comments are resolved?

alt-romes:
Yes, absolutely

PR #352, opened 2026-07-25
More assorted fixes for ghc-worker integration

Saizan:
Tried to make each commit self-explanatory. Let me know if this format works for a few small changes or if I should open separate PRs.

Until I'm back I'll still keep an eye on this PR and will be able to fix small things if needed to merge.

### DIFF
--- a/haskell-debugger/GHC/Debugger/Session.hs
+++ b/haskell-debugger/GHC/Debugger/Session.hs
@@ -31,7 +31,7 @@ module GHC.Debugger.Session (
   makeDynFlagsAbsoluteOverall,
   resumeExec,
   exposeModGraphUnitsInInteractiveGhcDebuggerUnit,
-  graphUnits,
+  graphsUnits,
   compileModuleWithDepsInHpt,
   home_unit_dflags,
   packageImportDecl,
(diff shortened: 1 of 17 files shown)


haskell-debugger/GHC/Debugger/Session.hs
```diff
@@ -270,16 +270,26 @@ addInteractiveGhcDebuggerUnit exposed env = do
...
          InstantiationNode uid _ -> Just uid
          LinkNode _ _ -> Nothing
+    mod_graph = hsc_mod_graph env
+    hug = hsc_HUG env
+    -- external deps do not always end up as UnitNodes,
+    -- so we look at the preloadUnits of home units as well.
+    explicit =
+      [ uid
+      | (_,hue) <- unitEnv_assocs hug
+      , uid <- GHC.preloadUnits $ homeUnitEnv_units hue ]
```
alt-romes: here when you say "as well" I think you mean because we get the preloadUnits and the UnitNodes out of the mod_graph and call those external deps? It would be better to clarify what is meant by "as well"

alt-romes (approved):
lgtm thanks!

haskell-debugger/GHC/Debugger/Session.hs
Saizan: I'm using external deps to mean deps that are not home units, I assumed that as common terminology
 
so by "as well" I mean we also look for them in preloadUnits

I'll clarify the comment

alt-romes: that’s much better now! land it

PR #355, opened 2026-08-19
[Fix #160] Use bcoArgs to augment getVariables

### LINKED ISSUE #160: Per-stack-frame variables
With #158, we already have a good callstack comprised of both IPE frames and `RetBCO` frames (aka interpreter continuation frames).
(Note: Tail calls lead to sometimes confusing callstacks, especially when the user doesn't have tail calls in mind. This is just the debugger observing how the execution actually works -- we can't display frames that don't exist)

This ticket tracks the improvement of callstacks by making each stack frame display the variables bound in the scope of that frame. Currently, we always display the bindings of the last frame.

To achieve this, we want to combine information from 2 locations.

1. The `ModBreaks` for the `InternalBreakpointId` associated with interpreter frames has a list of `OccName`s of source-level variables bound in that scope, in order.
2. The `StackFrame` from decoding the stack contains the `bcoArgs :: [StackField]` field which contains the arguments of a BCO found while decoding the stack.

We need to join this information, in a way similar to what `bindLocalsAtBreakpoint` does, to correctly report the variables in scope per-stack-frame

Saizan:
This is working on 9.14.1 atm.

TODO:
- [x] See if there's a sensible way to *temporarily* make stack frame vars available to the debug console.
    - not really, see issue #361 .
- [x] Add support for nightly.
    - it was easy to use the same stack inspection, punted the rest to issue #362  
- [x] Should we cache the binding of the stack frame vars?
   - dap-clients (vscode at least) already cache the request, so doesn't seem necessary.
   
There's a couple more TODOs in the diff, I would like some input there.


haskell-debugger.cabal
```diff
@@ -158,11 +160,16 @@ library
 
                       haskell-debugger-view >= 0.2 && < 1.0,
                       ghc-stack-annotations >=0.1 && <0.2,
+                      rts,
 
     if !os(windows)
         build-depends: unix >= 2.8.6 && < 2.9,
-
+    if impl(ghc>=10.1)
+        cc-options: -DFEW_T
```
alt-romes: What does this stand for?

haskell-debugger/GHC/Debugger/Session/Builtin.hs
```diff
@@ -173,3 +203,10 @@ debuggerViewTextContents = stringToStringBuffer $(embedStringFile =<< makeRelati
 -- | GHC.Debugger.View.ByteString
 debuggerViewByteStringContents :: StringBuffer
 debuggerViewByteStringContents = stringToStringBuffer $(embedStringFile =<< makeRelativeToProject "haskell-debugger-view/src/GHC/Debugger/View/ByteString.hs")
+
+debuggerRuntimeFFIInspectModName :: ModuleName
+debuggerRuntimeFFIInspectModName = mkModuleName "GHC.Debugger.Runtime.FFIInspect"
+
+-- | The contents of GHC.Debugger.Runtime.FFIInspect in memory
+debuggerRuntimeFFIInspectContents :: StringBuffer
+debuggerRuntimeFFIInspectContents = stringToStringBuffer $(embedStringFile =<< makeRelativeToProject "haskell-debugger/GHC/Debugger/Runtime/FFIInspect.hs")
```
alt-romes, line 218: So this module is executed both as part of the host debugger but also in the debuggee program? It would be good to state this in the comment, and the reason why both us and the ext interp need to execute things from this module.

Or, it might be worth to split the parts which are meant to be executed in the debuggee from the others. That would make it less confusing when the module is being used/run.

If that's better in your opinion, then we should also extract the module into a separate folder other than `haskell-debugger/...`, to make it clear it's only meant to be loaded at runtime.

haskell-debugger/GHC/Debugger/Session/Builtin.hs
```diff
@@ -96,9 +102,32 @@ addInMemoryHsDebuggerViewUnit
...
+hsDebuggerFFIInspectUnitId :: UnitId
+hsDebuggerFFIInspectUnitId = stringToUnitId "haskell-debugger-ffi-inspect"
+
+addInMemoryFFIInspectUnit :: GhcMonad m => [UnitId] -> DynFlags -> m UnitId
+addInMemoryFFIInspectUnit deps dflags = do
+  addInMemoryUnit
+    hsDebuggerFFIInspectUnitId
+    (coerce hsDebuggerFFIInspectUnitId)
+    deps dflags
+  return hsDebuggerFFIInspectUnitId
```
alt-romes, line 124: In a way similar to how I've moved the code which doesn't use the custom external commands into the `GHC.Debugger.Runtime.Interpreter.Legacy`, it would be good if this code lived in a module which is also Only ever included in the exposed modules when `impl(ghc <= 9.14.1)`. If not, at least add CPP to make it obligatory it isn't used for 9.14.2.

This is important because we hope to drop this code completely when 9.14.2 is well-established (e.g. by the time 9.14.5 is out), but also because it makes it much less confusing if we can tell just from the structure what is used in 9.14.1 vs 9.14.2, because the code paths are just very different.

I think this might be solved by not having a separate unit necessarily for the code that you need to execute. What is needed from that module at debuggee vs debugger specifically, and could the debuggee parts be inline code like we have for many things?

haskell-debugger.cabal
```diff
@@ -158,11 +160,16 @@ library
...
     hs-source-dirs:   haskell-debugger
+    include-dirs:     cbits
+    c-sources:        cbits/stack.c
+    install-includes: stack.h
```
alt-romes, line 182: Is this code ever run in the debugger process? If not, then we should perhaps avoid it being here.

haskell-debugger.cabal
```diff
@@ -99,6 +99,8 @@ library
                       GHC.Debugger.Runtime.Compile,
                       GHC.Debugger.Runtime.Compile.Cache,
 
+                      GHC.Debugger.Runtime.FFIInspect,
```
alt-romes, line 108: I think this is only meant to be used in the debuggee process, so it shouldn't be an exposed module of the debugger? I think.

haskell-debugger.cabal
alt-romes: Answer: yes, when using the custom external interpreter commands we call these bcoOffsets functions directly (the debuggee process is our custom external interpreter which is the same executable as the debugger) without having to load them dynamically.

haskell-debugger.cabal
alt-romes: Answer: the point is it is used in the external commands when this executable serves as the external interpreter

cbits/stack.c
```diff
@@ -0,0 +1,264 @@
...
+#include "Rts.h"
+#include <stddef.h>
+#include <string.h>
+
+/*
+Note [Ask the RTS for memory layout of AP_STACK]
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Goal: find values for breakpoint variables in RET_BCO continuation/apply stack frames.
+Challenge: breakpoint metadata contains only their offsets in AP_STACK closures, which get built only when evaluation stops at the breakpoint: typically a copy of both the continuation and return frames. When we only have the RET_BCO frame in hand we need our own offset to adjust the AP_STACK ones.
```
alt-romes: I would lead the note as follows, replacing what you have until the first `## RET_BCO frames reminder`:

The bytecode compiler in ghc currently assigns to each breakpoint captured variable an offset to read that variable off of a pointer into an AP_STACK constructed when we stop at that breakpoint. However, when we want to read the variables of frames from breakpoints of *other* frames on the stack, that offset can't be used to index other frame's free variables, because it is meant specifically as an offset into the AP_STACK constructed when we stop at a breakpoint.

While ghc doesn't provide more information regarding a variable offset into a frame's free vars on the stack, rather than into the stopped-at-breakpoint-AP_STACK, we reverse engineer the offset into the free variables from the offset we are given.

There are two main cases we need to distinguish to reverse engineer the offset into a frame's variables correctly:

- The frame is a function or case continuation (type: RET_BCO), and the offset into the offset is fixed
- The frame is a case continuation for an unboxed tuple-scrutinee (type: RET_BCO, header == &stg_ctoi_t_info), and the offset into the offset depends on the tuple size which we can read from the stack

haskell-debugger.cabal
alt-romes: FEW_T as in fewer `ctoi_tN` constructors... ehh add an inline comment saying something like

-- from 10.1, we have much fewer `stg_ctoi_t<N>_info` hardcoded frames

cbits/stack.c
```diff
@@ -0,0 +1,264 @@
...
+#undef ENT
+
+#define BCO_FRAME_TABLE_LEN \
+    (sizeof(bco_frame_table) / sizeof(bco_frame_table[0]))
+
+// Checks the frames listed are actually RET_BCO and that they are distinct.
+bool stack_bco_frame_selftest(void)
+{
+    bool flag = true;
+    errorBelch("stack_bco_frame_selftest\n");
```
alt-romes: hmm?

cbits/stack.c
```diff
@@ -0,0 +1,264 @@
...
+};
+
+#undef ENT_T
+#undef ENT
+
+#define BCO_FRAME_TABLE_LEN \
+    (sizeof(bco_frame_table) / sizeof(bco_frame_table[0]))
+
+// Checks the frames listed are actually RET_BCO and that they are distinct.
+bool stack_bco_frame_selftest(void)
```
alt-romes: nice

cbits/stack.c
```diff
@@ -0,0 +1,264 @@
...
+{
+    bool flag = true;
+    errorBelch("stack_bco_frame_selftest\n");
+    for (size_t i = 0; i < BCO_FRAME_TABLE_LEN; i++) {
+        const StgInfoTable *it =
+            INFO_PTR_TO_STRUCT((const StgInfoTable *)bco_frame_table[i].info);
```
alt-romes, line 182: Why `INFO_PTR_TO_STRUCT`? I thought we were already writing the info table pointer for the RET_BCO frames into the array directly

alt-romes: bc with tables next to code the info table pointer points to the code directly rather than the info, yeah

cbits/stack.c
```diff
@@ -0,0 +1,264 @@
...
+        }
+    }
+    return flag;
+}
+
+// Returns BcoFrameEntry for given stack frame.
+// In can be of the entries in bco_frame_table or one of the BCO_FRAME_NOT_BCO, BCO_FRAME_UNKNOWN kinds.
+BcoFrameEntry stack_bco_frame_entry(const StgClosure *frame)
+{
+    const StgInfoTable *info = frame->header.info;
```
alt-romes, line 226: Shouldn't this be `info = get_itbl(frame)`? then below `info->type`?

alt-romes: as is, with tables next to code, `frame->header.info` points to the code rather than the info table info

alt-romes: ah, but below you compare the info table addresses directly with the ones in the BCO_FRAME_TABLE. fine, it all adds up. keep it as is!

cbits/stack.c
```diff
@@ -0,0 +1,264 @@
...
+StgInt bco_frame_args_offset(StgClosure * frame)
+{
+    BcoFrameEntry frame_entry = stack_bco_frame_entry(frame);
+
+    switch (frame_entry.kind){
+        case BCO_FRAME_NOT_BCO:
+        case BCO_FRAME_UNKNOWN:
+            return -1;
+        case BCO_FRAME_APPLY_INTERP:
+            return 2;
```
alt-romes: `return 2; // constant offset`

cbits/stack.c
```diff
@@ -0,0 +1,264 @@
...
+
+// Mapping from continuation/apply frame to its return frame (info pointers) and other metadata.
+static const BcoFrameEntry bco_frame_table[] = {
+    ENT(stg_apply_interp, BCO_FRAME_APPLY_INTERP, NULL) //frame used to suspend run_BCO, not to wait for a return.
+    ENT(stg_ctoi_R1p,     BCO_FRAME_CTOI_R1P,     &stg_ret_p_info)
+    ENT(stg_ctoi_R1n,     BCO_FRAME_CTOI_R1N,     &stg_ret_n_info)
+    ENT(stg_ctoi_F1,      BCO_FRAME_CTOI_F1,      &stg_ret_f_info)
+    ENT(stg_ctoi_D1,      BCO_FRAME_CTOI_D1,      &stg_ret_d_info)
+    ENT(stg_ctoi_L1,      BCO_FRAME_CTOI_L1,      &stg_ret_l_info)
+    ENT(stg_ctoi_V,       BCO_FRAME_CTOI_V,       &stg_ret_v_info)
```
alt-romes, line 149: shouldn't the ret_bco_offset be `1` in these cases? you currently have them as `0` which you say means "not a RET_BCO" despite these being RET_BCO in fact

cbits/stack.c
```diff
@@ -0,0 +1,264 @@
...
+            return 2;
+        default:
+            const void * ret_frame_info = frame_entry.ret_info;
+            StgBCO * ret_bco;
+            if(frame_entry.ret_bco_offset > 0){
+                ret_bco = (StgBCO *)((StgPtr)frame)[frame_entry.ret_bco_offset];
+            }
+            else{
+                ret_bco = NULL;
+            }
```
alt-romes: and here you are returning NULL for all the `ctoi_v,r,p,f,...` because those offsets are set to 0 at the moment. I think it ought to be 1.

cbits/stack.c
```diff
@@ -0,0 +1,264 @@
...
+            // We create a mocked return frame `stub` that contains the right header.
+            // The second word is only relevant for RET_BCO frames, and will be ignored for the others.
+            StgWord stub[2] = { (StgWord)ret_frame_info, (StgWord)ret_bco };
```
alt-romes: ah, so this `ret_bco` isn't a ret_bco at all! it's a `tuple_BCO`!!

all things referring `ret_bco` need to be renamed to `tuple_BCO`. `ret_bco` is understood to be the BCO that will execute when we return to this frame...

and that's why we only have `3` for the tuple ret bcos and 0 for th others.

cbits/stack.c
```diff
@@ -0,0 +1,264 @@
...
+            const void * ret_frame_info = frame_entry.ret_info;
+            StgBCO * ret_bco;
+            if(frame_entry.ret_bco_offset > 0){
+                ret_bco = (StgBCO *)((StgPtr)frame)[frame_entry.ret_bco_offset];
+            }
+            else{
+                ret_bco = NULL;
+            }
+            // We create a mocked return frame `stub` that contains the right header.
+            // The second word is only relevant for RET_BCO frames, and will be ignored for the others.
```
alt-romes: "The second word is only relevant for RET_BCO frames" is also not correct. Prefer:

"The second word is only used to determine the stack frame size for `stg_ret_t_info` frames, whose first word is the `tuple_BCO` which contains dynamic information about the size of the tuple being returned"

alt-romes (commented):
I'm still going through the patch, but I think there are a few misnomers which complicated reasoning a lot! I've pointed them all out.

cbits/stack.c
```diff
@@ -0,0 +1,264 @@
...
+            StgWord stub[2] = { (StgWord)ret_frame_info, (StgWord)ret_bco };
+            return (StgInt)stack_frame_sizeW((StgClosure *)stub) + 2;
+    }
+
+}
+
+
+StgInt bco_args_offset(StgStack* stack, StgWord index){
```
alt-romes: `..., StgWord frame_index)` would be clearer

haskell-debugger/GHC/Debugger/Session/Builtin.hs
Saizan: it's currently always used by the "debuggee", either loaded in memory for the Legacy path or directly imported for the implementation of Custom commands.

the only exception is the selftest which is used in the testsuite.

cbits/stack.c
alt-romes: Answer: no, the `ret_bco_offset` is actually `tuple_bco_offset` and should be renamed as such. Indeed, these frames don't have a tuple bco.

cbits/stack.c
alt-romes: Answer: see above, this is `tuple_bco`, not `ret_bco`.

haskell-debugger/GHC/Debugger/Interface/Messages.hs
```diff
@@ -340,7 +342,7 @@ data DbgStackFrame
     -- ^ Title of stack frame
     , sourceSpan :: SourceSpan
     -- ^ Source span for this stack frame
-    , breakId :: Maybe GHC.InternalBreakpointId
+    , breakId :: Maybe (GHC.InternalBreakpointId, Maybe (NoShow (ForeignRef [Heap.StackField]),Word))
     -- ^ Is this a BCO continuation frame with a breakpoint?
     -- If yes, we can leverage the breakpoint info to report scopes.
```
alt-romes: Expand the haddock to say what is the `ForeignRef [StackField], Word`

cbits/stack.c
Saizan: ah, I didn't realize that `ret_bco_offset` is ambiguous.

it's meant to be `offset of the return frame bco` like where to find tuple_bco inside the stg_ctoi_t[N] frames. So no, it shouldn't be 1, because it's not the offset of *this frame* bco.

`bco_of_return_frame_offset` seems better? I will also mention `tuple_bco` in the comment.

cbits/stack.c
Saizan: NULL is correct, the *return* frames for those ctoi_ ones are not RET_BCO themselves.

cbits/stack.c
Saizan: The comment is referring to whether the **return** frame is a RET_BCO as well or not.

haskell-debugger/GHC/Debugger/Runtime/Interpreter/Types.hs
```diff
@@ -20,11 +22,11 @@ data ThreadInfo ref = ThreadInfo
...
-data StackFrameInfo
+data StackFrameInfo ref
   -- | Information derived from an IPE entry
   = StackFrameIPEInfo !InfoProv
   -- | User-defined Stack Frame annotation
   | StackFrameAnnotation !(Maybe Stack.SrcLoc) !String
   -- | Information derived from a continuation BCO breakpoint info.
-  | StackFrameBreakpointInfo !InternalBreakpointId
-  deriving (Generic)
+  | StackFrameBreakpointInfo !InternalBreakpointId !(Maybe (NoShow (ref [Heap.StackField]),Word))
```
alt-romes: Could you create a datatype (and re-use it in the other place you inserted this ref `Maybe [StackField], Word`) that makes it clearer what is this pair of things?

cbits/stack.c
Saizan: I figured there could be more return frames that are RET_BCO than the tuple_bco ones, that's why I kept it to a more generic name.

FYI, the `ret` here is the one from the `stg_ret_$foo_info` naming convention, just slipped my mind that next to bco it's going to look like the closure type.

cbits/stack.c
alt-romes: Ah, ahhhhhhhh....

haskell-debugger/GHC/Debugger/Stopped.hs
```diff
@@ -300,14 +323,17 @@ getVariables threadId frameIx vk = do
...
 
     -- (VARR)(a) from here onwards
-
     LocalVariables -> fmap VariableFields $ do
-      -- bindLocalsAtBreakpoint hsc_env (GHC.resumeApStack r) (GHC.resumeSpan r) (GHC.resumeBreakpointId r)
-      mapM (tyThingToVarInfo fam_envs) =<< GHC.getBindings
+      vars <- case frame.breakId of
+            -- top frame vars already bound when stopping
+            _ | frameIx == 0 -> GHC.getBindings
+            Nothing -> pure [] -- TODO: we could output bcoArgs with fresh ids?
```
alt-romes: A good idea

haskell-debugger/GHC/Debugger/Stopped.hs
```diff
@@ -344,6 +370,170 @@ getVariables threadId frameIx vk = do
 
     -- Couldn't find ibi or frame
     _otherwise -> pure (VariableFields [])
+  where
+   getStackFrameBindings frame (ibi,Nothing) = do
```
alt-romes: All the new code required to implement this function should be moved to a new module, something like `GHC.Debugger.Stopped.Frames`

alt-romes (commented):
Excellent work!

cbits/stack.c
alt-romes: it's very confusing because there is a `stg_ret_` "return frame" and an `stg_ctoi_` "frame you return to", but the latter is also called way too often the "return frame" bc. term overloading

cbits/stack.c
alt-romes: Yeah, see above. "*return* frames" is too overloaded a term. I see what you mean now, but "return frame" is too often the frame we return to.

```c
/* -----------------------------------------------------------------------------
   Info tables for returning values of various types.  These are used
   when we want to push a frame on the stack that will return a value
   to the frame underneath it.
   -------------------------------------------------------------------------- */
```

there's this comment above the `stg_ret_*` entries. maybe we can call them "return_a_value_frame", that is less ambiguous.

cbits/stack.c
alt-romes: It's very hard to get these names right because of all the overloading. We do need something that is clearer, since this is quite tricky code.

alt-romes:
Excellent work addressing the review! I'm much happier with it now.

Small thing: could you make the export list of `module GHC.Debugger.Stopped.Frames where` explicit? IIUC not much should be exported out of the module

cbits/stack.c
```diff
@@ -0,0 +1,311 @@
...
+#define BCO_FRAME_IS_CTOI_TN(k) \
+    ((k) >= BCO_FRAME_CTOI_TN_BASE && \
+     (k) <= BCO_FRAME_CTOI_TN_BASE + BCO_FRAME_CTOI_TN_MAX)
+
+// BcoFrameEntry represents continuation/apply frames we care about.
+typedef struct {
+    const void *info;           // the stg_$name_info pointer of the frame.
+    StgWord     kind;           // enum tag, only used for error reporting atm.
+    const char *name;           // readable name for debugging
+    const void *return_info;       // expected return frame info pointer.
```
alt-romes:
```suggestion
    const void *return_info;       // expected return frame info pointer (e.g. stg_ret_v_info, stg_ret_t_info)
```

alt-romes (approved):

alt-romes:
Land it when you're ready :+1:

PR #367, opened 2026-09-07
Minimize jobs on non-release PRs

Saizan:
Managed to include the release runners only on release workflow.

Also unrelated cleanups:
  - cleaned up testing w/ haskell-debugger-view
  - made it so the tag only triggers a dry-run and need a workflow call for actual release. 
  - includes DLL file for dap-server sublibrary.

Continues from #353


Saizan:
We don't yet check that compile-time ghc matches runtime ghc, as we only release for 9.14.1 for now.

alt-romes:
> We don't yet check that compile-time ghc matches runtime ghc, as we only release for 9.14.1 for now

I think we do, see 
```hs
-- | Fetch the runtime GHC version, according to hie-bios, and check it is the
-- same as the compile time GHC version
hieBiosRuntimeGhcVersion :: HIE.Cradle Void
                         -> ExceptT String IO Version
hieBiosRuntimeGhcVersion cradle = do
```

Saizan:
@alt-romes

Draft github release here https://github.com/well-typed/haskell-debugger/releases/tag/untagged-9d044f11a5324850247b

I tested the -ubuntu2024.tar.gz bindist, maybe you want to test the macOS one?

Otherwise PR is ready to review/merge.

alt-romes:
I just finished reviewing. Broadly looks alright to me but I don't understand all that much about GitHub actions. These Yamls suck.

Thanks for taking care of it.

In the draft CI, did you run all pipelines? I'm not sure if running all jobs is working that well (see https://github.com/well-typed/haskell-debugger/pull/371 which ran on the CI just now with all the jobs by default)

alt-romes (approved):
Let's see what happens!

Saizan:
@alt-romes CI is not in an ideal position on master, I did run the Release pipeline for the last commit here: https://github.com/well-typed/haskell-debugger/actions/runs/35098439734

alt-romes:
> @alt-romes
> 
> Draft github release here https://github.com/well-typed/haskell-debugger/releases/tag/untagged-9d044f11a5324850247b
> 
> I tested the -ubuntu2024.tar.gz bindist, maybe you want to test the macOS one?
> 
> Otherwise PR is ready to review/merge.

Ah, I understand what you mean now. Let me test the macOS release. Your draft pipeline did build them all.

Could you explain to me what happens in a draft release as opposed to a full release? Do we publish a package candidate on hackage? What about VScode Marketplace?

alt-romes (changes requested):
Ok, I've tried the draft release and I do have a comment:

The binary releases are very specifically tailored to GHCup. They are not like the binary distributions of say upstream GHC that we can download, `./configure`, and `make install` with any toolchain.

The binary releases are for an `hdb` specifically compiled by the GHC 9.14 shipped by non-vanilla GHCup and can't be used with any other GHC (like the one from vanilla GHC upstream, or one from nixpkgs), because the ABIs won't match.

That's all fine, since first class support for GHCup is important. But we should rename the binary distribution something else like ghcup-hdb instead of hdb.

Does that seem fair?

Saizan:
> Could you explain to me what happens in a draft release as opposed to a full release? Do we publish a package candidate on hackage? What about VScode Marketplace?

- If the version is not on hackage yet we publish a package candidate to Hackage
- The VSCode and the other marketplace do nothing (the actions would package the extension but we already do that separately).
- Github gives us a draft release, like I linked above, the logs have the link.

The Release job here was run with debug logging so you can see a lot of the details of what's going on.
https://github.com/well-typed/haskell-debugger/actions/runs/35098439734/job/104824963620

> That's all fine, since first class support for GHCup is important. But we should rename the binary distribution something else like ghcup-hdb instead of hdb.

And here you mean renaming the tarballs, not the exes in them, right?

Also, I suppose it would only be a bit more copy pasting to also have bindists that link against the vanilla release.

alt-romes:
> And here you mean renaming the tarballs, not the exes in them, right?

Yes, that's what I meant

> Also, I suppose it would only be a bit more copy pasting to also have bindists that link against the vanilla release.

Hmm, that's a potentially good idea. I wonder if GHCup would have any chance of selecting the appropriate `hdb` binary depending on whether the active GHC comes from the vanilla or default GHCup channel. How does HLS handle this? cc @fendor @hasufell

alt-romes (approved):

PR #368, opened 2026-09-09
[Fix #365] not in scope error for getClosureData crashes debugger

### LINKED ISSUE #365: Regression in debugging HLS
### Summary

It appears that launching `hdb` version 10488e6997de44d3d3eae84e402b554176c8052e on HLS codebase as described in the **Reproduction steps** section results in `hdb` being killed/shut down. Using instead _that_ version (_that_ = see screenshot at the bottom of this post) doesn't cause that behavior.

In neither case, though, is the 3rd breakpoint mentioned in the **Reproduction steps** section hit. This is probably a different issue to be opened?

Furthermore, I have the impression that 10488e6997de44d3d3eae84e402b554176c8052e is way slower than _that_ version. Maybe this is because of how I `cabal build`?

### Some surrounding facts

The workaround that @fendor shared in this https://github.com/well-typed/haskell-debugger/issues/320#issuecomment-4562806288 shouldn't be needed anymore with the `HEAD` of `cabal` (and indeed that's what I verified), because the https://github.com/haskell/cabal/issues/11850 was fixed by https://github.com/haskell/cabal/pull/12084.

### Reproduction steps

  1. Build 10488e6997de44d3d3eae84e402b554176c8052e, e.g. via
      ```bash
      cabal build all -j32
      ```
  2. obtain the path to the executable, e.g. via
      ```bash
      cabal list-bin  hdb
      ```
      which for me resulted in
      ```none
      /home/enrico/haskell-debugger/dist-newstyle/build/x86_64-linux/ghc-9.14.1/haskell-debugger-0.13.1.0/x/hdb/build/hdb/hdb
      ```
  3. paste that in the Haskell Debugger configuration; I use Vim + Vimspector, but I've also reproduced in VSCode by setting this:

     Image

  4. put breakpoints at these 3 locations:

      4.1 https://github.com/haskell/haskell-language-server/blob/eb3954e498bbe5b592bf87a73e0ee321b4e78723/plugins/hls-class-plugin/test/Main.hs#L24

      4.2. https://github.com/haskell/haskell-language-server/blob/eb3954e498bbe5b592bf87a73e0ee321b4e78723/plugins/hls-class-plugin/test/Main.hs#L47

      4.3. https://github.com/haskell/haskell-language-server/blob/eb3954e498bbe5b592bf87a73e0ee321b4e78723/hls-plugin-api/src/Ide/PluginUtils.hs#L124

  5. start the debugger while the cursor is in the former file, wait long enough, and the control will stop at the first breakpoint, as expected,

  6. press <kbd>F5</kbd>, and the cursor will stop on the second breakpoint, as expected,

  7. press <kbd>F5</kbd> once more, and the cursor **_won't_** stop on the third breakpoint, unexpectedly.

     **This last fact (breakpoint not hit) affects also _that_ version, not just 10488e6997de44d3d3eae84e402b554176c8052e.**

### Logs

Below are the contents of these 3 tabs from left to right:

Image

  - [OUTPUT](https://paste.c-net.org/BerthaRoughed)

  - [DEBUG CONSOLE](https://paste.c-net.org/CharlesGribbs)

  - TERMINAL:
    ```
    /home/enrico/haskell-debugger/dist-newstyle/build/x86_64-linux/ghc-9.14.1/haskell-debugger-0.13.1.0/x/hdb/build/hdb/hdb external-interpreter --port 35857
    class
    class
      code actions
        Produces addMinimalMethodPlaceholders code actions for one instance:                   OK (5.78s)
        Creates a placeholder for '==':                                                        *** Ignoring breakpoint
    *** Ignoring breakpoint
    *** Ignoring breakpoint
    *** Ignoring breakpoint
    *** Ignoring breakpoint
    *** Ignoring breakpoint
    *** Ignoring breakpoint
    *** Ignoring breakpoint
        Creates a placeholder for '==':                                                        OK (5.89s)
        Like previous one, but this file has no line terminator:                               *** Ignoring breakpoint
    *** Ignoring breakpoint
    *** Ignoring breakpoint
    *** Ignoring breakpoint
    *** Ignoring breakpoint
    *** Ignoring breakpoint
        Like previous one, but this file has no line terminator:                               FAIL (5.95s)
          Test output was different from 'plugins/hls-class-plugin/test/testdata/T1W.eq.expected.hs'. Output of ["git","-c","core.fileMode=false","diff","--no-index","--text","--exit-code","plugins/hls-class-plugin/test/testdata/T1W.eq.expected.hs","/tmp/T1W.eq.expected3676310-153.actual"]:
          diff --git a/plugins/hls-class-plugin/test/testdata/T1W.eq.expected.hs b/tmp/T1W.eq.expected3676310-153.actual
          index 37fb11727..1a16a4bc8 100644
          --- a/plugins/hls-class-plugin/test/testdata/T1W.eq.expected.hs
          +++ b/tmp/T1W.eq.expected3676310-153.actual
          @@ -3,4 +3,4 @@ module T1 where
           data X = X

           instance Eq X where
          -  (==) = _
          \ No newline at end of file
          +  (==) = 
          \ No newline at end of file

          Use -p '/Like previous one, but this file has no line terminator/' to rerun this test only.
        Creates a placeholder for '/=':                                                        OK (5.91s)
        Creates a placeholder for both '==' and '/=':                                          OK (5.94s)
        Creates a placeholder for 'fmap':                                                      *** Ignoring breakpoint
    *** Ignoring breakpoint
    ```

---

About the breakpoints, I've lied on the position of #2 and #3, because I've reproduced this in a working directory where I'm doing some work, but the because of what I explained in the _Summary_ section above, I'm relatively confident my edits are irrelevant to the issue.

Image

Saizan:
Fix #365.
TODO:
- [x] Made errors in parsers not fatal, only reported to client and logged.
- [x] Made setLineBuffering and UnpackStackFields into custom commands
- [x] Added GHC.Debugger.Runtime.Internal for helpers we need at runtime. See Note [debuggerInternal unit]


haskell-debugger/GHC/Debugger/Runtime/Internal.hs
alt-romes, line 1: should this most of this module be guarded as .Legacy in cabal under `impl(ghc <= 9.14.1)`? I would expect little of this to be needed in a Custom Commands World

Then this module can re-export that .Legacy module `#if MIN_VERSION_ghc`...

It would be clearer what parts are needed only because of the lack of custom commands.

alt-romes (commented):
Yes, I like the overall approach moving less thing from hardcoded strings into just a module which is compiled and loaded at runtime all at once.

I just did a first quick pass, I'll need to look again better.

alt-romes:
IIRC you previously introduced an in-memory-"FFI"-module. Could that be merged into this one perhaps?

Could we make somehow the distinction between a module that is used for the compilation of the debugger vs a module used only at runtime?

There is also one module which I think is all specified as a string or something which is also loaded at runtime. Or is just listed in `extra-source-files` and then inserted into the program using TH? In any case, it would be good to consolidate all these competing methods and make it clear the distinction between runtime vs compile-time modules (e.g. use a fully distinct module Namespace)

haskell-debugger/GHC/Debugger/Runtime/Internal.hs
Saizan: Actually the only thing to gate behind `ghc < 9.14.2` would be the re-export of getClosureData.

Everything else is used either by a custom command or unconditionally by an interpreted snipped that could not be made into a command.

Saizan:
@alt-romes

I don't really see "competing methods", tbf. The contents of all the modules you mentioned are bound to a StringBuffer CAF each via TH at compile time (the haskell-debugger-view ones need to be in extra-source-dirs so the files are in the haskell-debugger's sdist). 
They are all also built statically regardless of when they are used, which keeps the dev experience saner.
 
If you prefer, what we can do is have one module that exports all the things from `haskell-debugger` for sure needed at runtime. Under ghc < 9.14.2 the module will also re-export definitions that are from modules that are imported statically with ghc >= 9.14.2 (to implement custom commands), which means those re-exported modules will also be loaded at runtime (and must only depend on libraries shipped with ghc). 

The haskell-debugger-view in-memory unit has to stay separate though, because that one depends on the debuggee units (both as home unit dependencies and whether we load it or not).

More details on how things are used atm:
- In extra-source-files we include the modules (and .cabal? that does not seem useful) from haskell-debugger-view. That's only conditionally an in-memory unit, depending on the debuggee, and we want to release the library anyway.
- GHC.Debugger.Runtime.FFIInspect is used statically with Custom Commands (c.f. stackFrameInfo) and at runtime without. The unit is loaded at runtime only if ghc < 9.14.2. Hopefully this whole module goes away after we implement the better CgModInfo, which is why it's kept separate.
- GHC.Debugger.Runtime.Internal has two more functions like above (static with Custom Commands, runtime otherwise), and the rest are unconditionally used at Runtime. These definitions we will keep needing in the future.

haskell-debugger/GHC/Debugger/Runtime/Eval/RemoteExpr/Builtin.hs
```diff
@@ -66,7 +63,7 @@ decodeStackWithIpe = Remote.app $ Remote.var (mkModuleName "GHC.Internal.Stack.D
 
 -- | Remote 'GHC.Exts.Heap.getClosureData'
 getClosureData :: RemoteExpr StgStackClosure -> RemoteExpr (IO Closure)
-getClosureData = Remote.app $ Remote.var (mkModuleName "GHC.Exts.Heap") "getClosureData" ["GHC.Exts.LiftedRep", "_"]
```
alt-romes, line 69: why could the type arguments now be removed? A bit surprising, it used to require an explicit type application iirc. Well, no worries!

haskell-debugger/GHC/Debugger/Runtime/Interpreter/Legacy.hs
```diff
@@ -328,14 +336,19 @@ bcoBreakPointInfoParser = do
         liftIO $ fail "Failed to parse BCOClosure's BRK_FUN"
       Right r -> return r
 
+unpackStackFields :: ForeignRef [Stack.StackField] -> Maybe [Int] -> Debugger [ForeignHValue]
+unpackStackFields fldsRef mixs = runInternal $ do
+  (expectRight =<<) $ Remote.evalIOList $
```
alt-romes, line 341: should this be `liftDebuggerOrFail` `NonFatal` style?

haskell-debugger/GHC/Debugger/Runtime/Eval/RemoteExpr/Builtin.hs
alt-romes: IIRC as well you had initially exported `getClosureData` from the internal runtime module, why not anymore?

haskell-debugger/GHC/Debugger/Runtime/Interpreter/Custom.hs
```diff
@@ -236,6 +244,40 @@ exceptionInfo prefix se'@(SomeException exc) =
...
+  ExternalInterp{}
+    -- Use interpCmd to send the command as a custom message to external process.
+    -- The custom message will be processed according to the custom command
+    -- handlers registered with `iservWithCustom`
+    | let payload = encodePayload (Some @Bin.Binary command) -> do
+      respBytes <- interpCmd interp (CustomMessage dbgInterpCmdTag payload)
+      case decodePayload respBytes of
+        Left err -> fail err
+        Right r  -> pure r
+
```
alt-romes: I dislike that this is moved here.

The GHC.Debugger.Runtime.Interpreter.Custom module currently only contained functions which are meant to be executed by the interpreter (typically the external one). But this `interpDbgCmd` is executed by the debugger.

My suggestion is:

1. Add some haddocks to the module header saying the functions on this module are meant to be executed by the interpreter at runtime and never by the debugger directly, which should always go through `interpDbgCmd`.
2. Move `interpDbgCmd` somewhere else. An easy solution would be yet another module like "Runtime.Interpreter.Cmd". I've tried thinking about a few more solutions that don't involve creating more modules, but I did not find a satisfactory one

haskell-debugger/GHC/Debugger/Runtime/Internal.hs
```diff
@@ -0,0 +1,55 @@
+-- | A module providing various remote external variables which we want
+-- available in any session. It gets compiled in-memory at the start of one.
+--
+-- See Note [debuggerInternal unit].
+{-# LANGUAGE LambdaCase #-}
+module GHC.Debugger.Runtime.Internal
+  ( module GHC.Debugger.Runtime.Internal
+  , GHC.evalWrapper
```
alt-romes, line 8: having the evalwrapper here will be really useful in the near future, e.g. to label the debuggee thread something other than "GHCi sandbox" before executing the main program (which might override it again, NP)

haskell-debugger/GHC/Debugger/Runtime/Eval/RemoteExpr/Builtin.hs
```diff
@@ -111,3 +108,8 @@ displayStackAnnotationShort = Remote.app $ Remote.var (mkModuleName "GHC.Interna
 
 stackAnnotationSourceLocation :: RemoteExpr SomeStackAnnotation -> RemoteExpr (Maybe SrcLoc)
 stackAnnotationSourceLocation = Remote.app $ Remote.var (mkModuleName "GHC.Internal.Stack.Annotation") "stackAnnotationSourceLocation" ["SGHC.Stack.Annotation.Experimental.omeStackAnnotation"]
+
+unpackStackFields :: Remote.RemoteExpr ([GHC.StackField] -> Maybe [Int] -> IO [HValue])
+unpackStackFields = Remote.var debuggerRuntimeInternalModName "unpackStackFields" []
```
alt-romes, line 113: this could be moved to the `.Legacy` module where this is needed, similarly to bcoArgsOffsets which is also defined just there.

haskell-debugger/GHC/Debugger/Runtime/Internal.hs
```diff
@@ -0,0 +1,55 @@
...
+-- See Note [Forcing debuggee's thunks].
+unpackStackField :: StackField -> IO HValue
+unpackStackField x = case x of
+  (StackBox (Box a)) -> returnIO (HValue a)
+  (StackWord w) -> returnIO (HValue (Unsafe.Coerce.unsafeCoerce w))
+
+unpackStackFields :: [StackField] -> Maybe [Int] -> IO [HValue]
+unpackStackFields fs = \case
+  Nothing -> mapM unpackStackField fs
+  Just xs -> flip mapM xs $ \ i ->
```
alt-romes, line 44: it's unfortunate that these definitions have to always be defined here (I suspect CPP wouldn't work for a compiled in memory module? or would it?)

In any case, could you make a section for these two which says something about "only needed for 9.14.1 legacy; subsumed by custom external interpreter command on 9.14.2"

alt-romes: Ah, nevermind.

```
    mapM mkRemoteRef =<< Internal.unpackStackFields flds mixs
```

The implementation of the custom command is calling this function from this module.

haskell-debugger/GHC/Debugger/Breakpoint.hs
```diff
@@ -251,13 +251,15 @@ condBreakEnableStatus hitCount condition = do
 --   Braces are preserved if escaped with a backslash. Some unescaped braces are
 --   fine: opening braces in antiquotations and closing braces outside of them.
 logMessageExpression :: String -> String
-logMessageExpression tmpl = apply "Prelude.putStrLn" $ apply "Prelude.concat" $ parts
+-- TODO: user Internal.
```
alt-romes: what does this todo mean? has it been finished? we are using `internal` below

haskell-debugger/GHC/Debugger/Monad.hs
```diff
@@ -394,14 +402,16 @@ runDebuggerAction l rootDir extraGhcArgs conf loadHomeUnit (Debugger action)
...
-            ; System.IO.hSetBuffering System.IO.stderr System.IO.LineBuffering }
-        """
+      setLineBuffering <- do
+         -- We do the #if dance here to avoid import cycles
+#if MIN_VERSION_ghc(9,14,2)
+         pure $ Custom.setLineBuffering
+#else
+         code <- runInternal $ compileExprRemote $ moduleNameString debuggerRuntimeInternalModName ++ ".setLineBuffering"
+         pure $ flip evalIO code
+#endif
```
alt-romes: perhaps one solution to the set buffering problem is to remove it wholesale as a custom command. Just always go through compileExprRemote "GHC.Debugger.Runtime.Internal.setLineBuffering", like we did before, but without the name conflicts. That would be a good solution to the cycles and doesn't seem too bad to me.

alt-romes (commented):
Excellent work!

I've got a few comments on a few things, but it's almost ready to land. Minimal changes needed.

haskell-debugger/GHC/Debugger/Runtime/Interpreter/Custom.hs
Saizan: So the distinction is that before we only had functions that were implementing Custom commands while now we also have ones that call those commands?

haskell-debugger/GHC/Debugger/Monad.hs
Saizan: I see you dislike CPP more than you want to use custom commands.

Indeed it should be fine to keep the `compileExprRemote` route

haskell-debugger/GHC/Debugger/Runtime/Eval/RemoteExpr/Builtin.hs
Saizan: Oops, good catch, I guess we use it in a context where the types can be inferred so no tests failed.

I went back and forth on this one because it's only needed by .Legacy, unlike everything else exported by GHC.Debugger.Runtime.Internal

haskell-debugger/GHC/Debugger/Runtime/Interpreter/Legacy.hs
Saizan: Not sure I get what you mean, but expectRight already makes this a NonFatal error if it happens

alt-romes: Ok never mind then 👍

haskell-debugger/GHC/Debugger/Runtime/Interpreter/Custom.hs
alt-romes: > So the distinction is that before we only had functions that were implementing Custom commands while now we also have ones that call those commands?

Yes, exactly. Previously the functions were only implementing Custom commands (and would be called in the *interpreter process*) whereas with this addition they now mix code that is meant to be run in the *interpreter process* and the *debugger process*

haskell-debugger/GHC/Debugger/Monad.hs
alt-romes: I don't mind CPP that much, but it is becoming confusing in the code base with all the interpreter vs debugger vs custom commands vs remoteeval vs various GHC versions what is happening.

My motivation for removing this `Custom.setLineBuffering` is mostly to kill the module cycle so that we can put `interpDbgCmd` back somewhere sensible.

haskell-debugger/GHC/Debugger/Runtime/Interpreter/Custom.hs
Saizan: Ok, I'll add the comment and land it.

alt-romes (approved):

PR #371, opened 2026-09-15
First pass for multi-threaded debugging

alt-romes:
- **Receive externalInterpreterArgs in launch.json**
- **multi-threaded: Display all threads**
- **Multi-threaded debugging first steps**
- **Refactor that cleans up module structure and explicit exports**
- **threaded: Eval request in thread context**
- **Make the testsuite driver multi-thread capable**
- **interactive: DoEval request doesn't change runCurrentThread**
- **interactive: Stepping shouldn't say Stopped at Breakpoint**


alt-romes:
Depends on a new `dap` version being released and merging into GHC upstream https://gitlab.haskell.org/ghc/ghc/-/merge_requests/16630

alt-romes:
The new `dap` version has been released, but we still need to upstream all multi-threaded debugging changes to GHC before this "just works". It already works with 9.14 though.

I'm a bit unsure what to do here. For one, it'd be important to land these foundations of multi-threaded to avoid further conflicts. I've also renamed a few things.

On the other hand, the nightly doesn't build currently. Perhaps I can tighten the bounds further to only use the multi-threaded features in an even more recent GHC version than the nightly?

I'll try that.

haskell-debugger/GHC/Debugger.hs
```diff
@@ -17,20 +27,56 @@ import GHC.Debugger.Interface.Messages
...
+
+  ------------------------------------------------------------------------------
+  -- GHC.Debugger.Run
+  ------------------------------------------------------------------------------
+
+  DoEval mfid exp_s -> DidEval <$>
+    Run.doEvalCommand mfid exp_s
+
+  -- TODO: We shouldn't block waiting for the result of these operations, because that means we can never resume/step two threads simultaneously. Recall
+  -- Recall we take things to evaluate from the message queue, but we execute them serially. Here we should do something like `forkIO $ reply`
```
Saizan:
```suggestion
  -- TODO: We shouldn't block waiting for the result of these operations, because that means we can never resume/step two threads simultaneously.
  -- Recall we take things to evaluate from the message queue, but we execute them serially. Here we should do something like `forkIO $ reply`
```

haskell-debugger/GHC/Debugger.hs
```diff
@@ -17,20 +27,56 @@ import GHC.Debugger.Interface.Messages
...
+  -- We need an async model that sets them off running and replies once the answer comes back
+  -- We should just do this in the dap side thread which is running serially
+  --
+  -- Even more generally, I think when we read an handleExecResult we should
+  -- probably clear the MVar as soon as possible so we can receive as many paused
+  -- thread hits as possible and emit thread paused events.
```
Saizan, line 77: Currently all interactions with the GHC session happen serially in debuggerThread, right?

Would be good to discuss what the new thread architecture should be, I couldn't quite follow what this TODO suggests. Can happen after we merge this PR of course.

haskell-debugger/GHC/Debugger/Monad.hs
```diff
@@ -107,9 +108,12 @@ data DebuggerState = DebuggerState
       , rtinstancesCache  :: IORef RuntimeInstancesCache
       -- ^ RuntimeInstancesCache
 
-      , threadMap         :: IORef TM.ThreadMap
+      , threadMap         :: IORef (TM.ThreadMap (ForeignRef ThreadId))
       -- ^ 'ThreadMap' for threads spawned by the debuggee
 
+      , threadResumeMap   :: IORef (TM.ThreadMap Resume)
+      -- ^ If a thread is currently stopped on a breakpoint, this map will contain
```
Saizan: ..will contain what? Such suspense!

haskell-debugger/GHC/Debugger/Stopped/Variables.hs
```diff
@@ -218,22 +266,18 @@ getFamInstEnvs' = do
...
       return emptyFamInstEnvs
 
-pprTerm :: Term -> SDoc
-pprTerm t = case t of
+--------------------------------------------------------------------------------
+-- Utils
+--------------------------------------------------------------------------------
+
+_ppr_term :: Term -> SDoc
+_ppr_term t = case t of
```
Saizan, line 274: What's the `_foo` name convention? private-ish like in python?

haskell-debugger/GHC/Debugger/Stopped/Variables.hs
```diff
@@ -22,11 +37,34 @@ import qualified GHC.Runtime.Heap.Inspect as GHCI
...
+--------------------------------------------------------------------------------
+
+-- | Get the value and type of a given 'Id' as rendered strings in 'VarInfo'.
+#ifdef GHC_HAS_MULTITHREADED_DBG
+idToVarInfo :: Id -> Debugger (Maybe VarInfo)
+idToVarInfo (GHC.AnId -> tt) = Just <$> do
+  fam_envs <- getFamInstEnvs'
+  tyThingToVarInfo fam_envs tt
+#else
+-- prefer idToVarInfo; this is no longer needed since ExecBreaks use Ids.
```
Saizan: Stale comment?

haskell-debugger/GHC/Debugger/Runtime/Interpreter/Custom.hs
```diff
@@ -138,6 +138,9 @@ stackFrameInfo ss (frameIx,(RetBCO{bco,bcoArgs}, _))
   = do
     xs <- NoShow <$> mkRemoteRef bcoArgs
     -- TODO: change ghc upstream so this offset is available from CgBreakInfo.
+    -- WARNING: This currently makes traversing the stack O(n^2), since
+    -- bcoArgsOffset needs to walk the stack until here and we do it for every
+    -- RetBCO frame. Changing it upstream would fix this asymptotic too.
```
Saizan, line 143: And walking down the stack in C is slow enough to matter? 

We could fix this by doing the offset lookup in bulk I suppose.

haskell-debugger/GHC/Debugger/Run/Resume.hs
```diff
@@ -0,0 +1,87 @@
...
+  -- The way it's done though also rollbacks any import statements that were
+  -- executed since the last ExecBreak. The only fix is to reimport everything
+  -- again.
+  --
+  -- Note: GHC.setContext recomputes the scope of the interactive imports from
+  -- scratch everytime. Considering `runDebugger` adds the whole home unit to
+  -- the interactive imports this might become a bottleneck. GHC does not keep
+  -- any reference to the scope containing just the imports, so we would have to
+  -- cache it ourselves (and then extend it with the cached scope of
+  -- ic_tythings, i.e. igre_prompt_env (c.f. replaceImportEnv)).
```
Saizan, line 85: Note to self: the imports added by `runDebugger` are instance-only ones (except Prelude), maybe we can avoid even those. The main purpose is bringing orphan DebugView instances in scope, but that could be accomplished by simply setting `ic_imports` directly (very hacky).

haskell-debugger/GHC/Debugger/Interface/Messages.hs
```diff
@@ -120,6 +125,36 @@ data Command
...
+--
+--   - When 'ResumeJustThisThread' is used, the debugger mode is toggled to
+--     non-stop-the-world, so, when that thread hits a breakpoint, it is only
+--     reported for that thread.
+--
+--   - When 'ResumeTheWorld' is used, the debugger mode is toggled to
+--     stop-the-world (regardless of threads having just been resumed with
+--     'ResumeJustThisThread'). From that point 'ResumeJustThisThread' is used
+--     again, all breakpoints hit pause all threads in addition to the one that
+--     hit the breakpoint.
```
Saizan: Should this be "From that point, until 'ResumeJustThisThread' is used again, ..."? Otherwise I don't follow

Saizan (approved):
Looks sensible and cleaned up! 

I noted some comment typos and some unclear spots.

haskell-debugger/GHC/Debugger.hs
alt-romes: Yes, exactly. There are two related things

- something like `resume` happens serially in debuggerThread, we always block waiting for 1 result (but we want to be able to resume 2 threads at the same time without seeing the result of either)
- when a breakpoint is hit by thread A, B, C, we're currently blocked on A until resuming and don't display that B and C are stopped to. this TODO is about clearing the status var as much as possible, so that we can emit three stopped events, for A, B, and C, in a row.

haskell-debugger/GHC/Debugger/Stopped/Variables.hs
alt-romes: This is dead code at the moment. Just thought might be useful if you want to trace something. I can kill it.

Saizan: Oh, indeed it seems useful to keep. Wanted to know that the `_` prefix signified though.

alt-romes: the prefix silents the unused warning

haskell-debugger/GHC/Debugger/Interface/Messages.hs
alt-romes: that's right! I've clarified it.

PR #374, opened 2026-09-17
Fix hackage artefacts

Saizan:


haskell-debugger.cabal
```diff
@@ -262,11 +262,11 @@ executable hdb
         time,
 
         directory >= 1.3.9 && < 1.4,
-        network >= 3.2.8,
-        network-run >= 0.4.4,
+        network >= 3.2.8 && < 3.3,
+        network-run >= 0.4.4 && < 0.5,
         async >= 2.2.5 && < 2.3,
         text >= 2.1 && < 2.3,
-        dap,
```
alt-romes, line 269: for what it's worth, you don't need to repeat the bounds on a different sub library, since the dependencies are solved whole-package (not per-sub-library)

alt-romes (approved):

haskell-debugger.cabal
Saizan: `cabal check` seemed to disagree, but icbw

Issue #220, opened 2026-03-13
Display debuggee threads


alt-romes:
Currently, only the thread which we're stopped at is reported in the returned Threads response.

However, the debuggee program may have multiple threads running and we definitely want to display all of them.
For threads which are stopped or blocked(e.g. on an MVar), we can inclusively display their stack frames.

This particular task has a handful of challenges:

- [x] We probably want to hide the threads labeled "IOManager" and "TimerManager"
- [x] We could try to relabel the main thread's label rather than the default `GHCi sandbox`. The user can overwrite it in their program afterwards.
- [x] ~~We need a strategy to update which threads are stopped and which are running~~ Left for other ticket 
    - One option is to figure out how to pause all threads whenever we hit a breakpoint, rather than just stopping one thread. This is what at least Python does.
    - Another option is to poll the status of the remaining threads in some interval in a background thread and send `Thread` events to notify the client that a thread is now stopped or otherwise.

PR #379, opened 2026-09-18
Hide interpreter threads and rename debuggee main

alt-romes:
- **Hide helper interpreter threads incl. interp server**
- **label the main debuggee thread**


alt-romes:
Screenshot 2026-09-18 at 15 09 11

This is what it looks like now, only the relevant threads including "Debuggee Main" (rather than "GHCi sandbox").

Not sure if Debuggee Main is the best name. Happy to take suggestions.

alt-romes:
Closes #220

Saizan (approved):

Issue #325, opened 2026-06-01
"Stop"-ing the dubugger before it finishes starting doesn't happen gracefully - Exceptions thrown in the console


Aster89:
I suspect there might be some other DAP non-compliance related to shutting down the debugger while it is starting.

I'm just reproducing @fendor 's [usecase](https://github.com/well-typed/haskell-debugger/issues/320#issuecomment-4562806288) from #320, and I'm doing it too in VSCode, hoping to make any investigation easier.

Here's a couple of scenarii where I launch the debugger and then click "Stop" (<kbd>Shift</kbd>+<kbd>F5</kbd>) at various moments/in various ways:

  - soon after having started it (say after a second)

    in the terminal window I see this:
    ```
    hdb: Uncaught exception ghc-internal:GHC.Internal.IO.Exception.IOException:

    Network.Socket.connect: <socket: 13>: does not exist (Connection refused)

    HasCallStack backtrace:
      ioError, called at ./Network/Socket/Internal.hs:224:5 in network-3.2.8.0-b69fc14ab47eb0f284d8e20e170e9ec0944dab54cb9a1b6cf51ab712a07bab38:Network.Socket.Internal
    ```
    in the output window, I see [this](https://www.dropbox.com/scl/fi/0m309cdzycawfk27blh7r/hdb-log-1?rlkey=0op7tpmoiiu02bnksguy5jpty&st=f662fft4&dl=0) (sorry, I used to upload to https://0x0.st/, but they've disabled it; what are you using nowadays? For now, I'm relying on my Dropbox)

  - after a few more seconds, but clicking "Stop" twice

    in the terminal window I see this:
    ```
    hdb: Uncaught exception ghc-internal:GHC.Internal.IO.Exception.IOException:

    GHCi.Message.readPipe: end of file

    HasCallStack backtrace:
      collectExceptionAnnotation, called at libraries/ghc-internal/src/GHC/Internal/Exception.hs:170:37 in ghc-internal:GHC.Internal.Exception
      toExceptionWithBacktrace, called at libraries/ghc-internal/src/GHC/Internal/Exception.hs:90:42 in ghc-internal:GHC.Internal.Exception
      throw, called at libraries/ghci/GHCi/Message.hs:721:16 in ghci-9.14.1-953d:GHCi.Message
    ```
[...]

PR #382, opened 2026-09-19
handle EOF uniformly and silently

Saizan:
We have a few forwarding threads and the client in TestDAP `forever` reading from an Handle. As there's no other stopping condition these are bound to raise EOF even when everything is working as expected, so the exception should be handled.

Some of these threads were not catching any exception which possibly made tests flakier than needed to be.

Others were catching `SomeException` and ignoring it without special casing AsyncExceptions, which could become a problem in itself.

Now these all use `silenceEOF :: Handle -> IO () -> IO ()` which ignores only EOF and only on the Handle we are trying to read from.

We should review if any of these should report the end of the stream higher up.

Also checked issue #325 to see if related and fixed the hClose problem from there.


haskell-debugger/GHC/Debugger/Utils.hs
```diff
@@ -31,15 +32,19 @@ import GHC.Debugger.Interface.Messages
 -- * Handle utils
 --------------------------------------------------------------------------------
 
+silenceEOF :: Handle -> IO () -> IO ()
+silenceEOF h m = do
+  m `catchNoPropagate`
+    \x@(ExceptionWithContext _ctx e) -> do
+      if isEOFError e && ioeGetHandle e == Just h
+        then return ()
```
fendor: do we want to maybe log this with `Debug` level or something?

Might still be relevant to know when this happens

hdb-dap/Development/Debug/Adapter/DAPDebuggee.hs
```diff
@@ -202,8 +202,8 @@ stdoutCaptureThread :: Maybe (Chan BS.ByteString) -> (DebugAdaptorCont () -> IO
 stdoutCaptureThread msyncOut withAdaptor = do
   tid <- myThreadId
   labelThread tid "Stdout Capture Thread"
-  withInterceptedStdout $ \_ interceptedStdout -> do
-    forever $ do
+  withInterceptedStdout $ \_ interceptedStdout -> silenceEOF interceptedStdout $ do
+    forever $ mask_ $ do
```
fendor: Might this deserve a comment?

It is not straight forward to me at all what  `mask_` achieves here.
At a first glance, it seems to me this thread is unkillable and relies on `hGetLine` throwing an EOF to cleanly exit.

hdb-dap/Development/Debug/Adapter/DAPDebuggee.hs
```diff
@@ -220,8 +220,8 @@ stderrCaptureThread :: Maybe (Chan BS.ByteString) -> (DebugAdaptorCont () -> IO
 stderrCaptureThread msyncErr withAdaptor = do
   tid <- myThreadId
   labelThread tid "Stderr Capture Thread"
-  withInterceptedStderr $ \_ interceptedStderr -> do
-    forever $ do
+  withInterceptedStderr $ \_ interceptedStderr -> silenceEOF interceptedStderr $ do
+    forever $ mask_ $ do
```
fendor: same here

alt-romes:
The plan you laid out in https://github.com/well-typed/haskell-debugger/issues/325#issuecomment-5799296480 is perfect. I think we should implement that.

hdb-dap/Development/Debug/Adapter/DAPDebuggee.hs
alt-romes: Since `mask` is inside the `forever`, it still has a chance to be killed every iteration. It means we never want to be killed by something between reading and sending.

But it would indeed be good to say why we care

hdb/Main.hs
```diff
@@ -127,10 +134,18 @@ main = do
...
-    withExternalInterpreterPort :: PortNumber -> (Handle -> IO a) -> IO a
-    withExternalInterpreterPort port k = do
-      bracket (mkHandleFromPortSock "127.0.0.1" port) hClose $ \ h -> do
+    withExternalInterpreterPort :: LogAction IO (WithSeverity String) -> PortNumber -> (Handle -> IO a) -> IO a
+    withExternalInterpreterPort l port k = do
+      bracket (mkHandleFromPortSock "127.0.0.1" port) hCloseGracefully $ \ h -> do
         annotateCallStackIO $ k h
+      where
+        hCloseGracefully h = do
+          catchNoPropagate (hClose h) $ \ x@(ExceptionWithContext _ e) ->
```
alt-romes: good catch (get it?)

haskell-debugger/GHC/Debugger/Monad.hs
```diff
@@ -380,10 +381,14 @@ cleanupInterp = do
...
           InterpRunning i  -> do
+            terminateProcess i.instProcess.interpHandle
+
             -- Can't use  `getProcessExitCode` because the interp process is
             -- not necessarily a child of this process (runInTerminal case).
             -- Just unconditionally try to send the message.
-            sendMessage i Shutdown
+
+            -- We control the process though, and we know it will not ignore sigTERM.
+            -- TODO: we could setup a side channel to confirm. (pidfd, known empty PIPE we could check for EOF, ..).
```
alt-romes: I recall fighting this for a while. IIRC `getProcessExitCode` itself is UB when used on non-child processes (like in the run in terminal case). As you say, we may be able to do this some other way.

But it also adds to the complexity. I liked your plan where we do best effort cancelation:

- SIGINT gives the debuggee thread a chance to exit (but the ext. interp shouldn't be killed)
- Send shutdown then, and expect EOF
- (Terminate) Destroy debug session and send terminated response, or reply error. (Disconnect) SIGKILL debuggee afterwards *anyway* and then destroy debug session; Always succeeds.

alt-romes (commented):
very good direction.

hdb-dap/Development/Debug/Adapter/DAPDebuggee.hs
fendor: yes, but it blocks on `hGetLine`, and I dont know whether that is an interruptible operation.

Saizan: My assumption was that any operation blocking via the RTS (as opposed to inside a FFI call) is interruptible.

The mask_ was in the pre-existing def of `forwardingThread` in our `Development.Debug.Adapter` module, and I figured we should make our forwarding threads more uniform. I should generalize forwardingThread so it can be used here too, not sure why I didn't yet.

On why it's important, I think we want these threads to be as transparent as possible so at a minimum we shouldn't drop a read line on the floor.

